PR #33, opened 2020-01-11
Fix two failing unittests for EmailAddressValidator using DNS

Zombaya:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description
I fixed the two tests who were responsible for the failed build.

#### 1. EmailAddressTest::testNoMxRecordARecordFallback() 
The domain `example.com` has a MX-record, making the test fail.
I replaced this by the domain `www.getlaminas.org` as it has no MX-record and is related to this package.

#### 2. EmailAddressTest::testUseMxCheckBasicValid() 
The domain `письмо.рф` is no longer registered.
I replaced this by the domain of the registrar of the `.рф`-tld.

### DIFF
--- a/test/EmailAddressTest.php
+++ b/test/EmailAddressTest.php
@@ -396,7 +396,7 @@ class EmailAddressTest extends TestCase
             $this->markTestSkipped('Testing MX records is not supported with this configuration');
         }
 
-        $email = 'good@example.com';
+        $email = 'good@www.getlaminas.org';
         $host = preg_replace('/.*@/', null, $email);
 
         //Assert that email host contains no MX records.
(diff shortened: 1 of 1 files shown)


Zombaya:
Well, the tests for 7.1,7.2 and 7.3 are passing again.  7.4 has other issues it seems.

Zombaya:
The reason of the deprecation-errors thrown in php 7.4 is because of an old version of phpunit.

The current version of phpunit (8), does not throw this deprecation-error, but will throw deprecation-warnings for phpunit itself since they will move to different testing-methods in phpunit 9.

See [Phpunit/phpunit #3728 ReflectionType::__toString() is deprecated in PHP 7.4 ](https://github.com/sebastianbergmann/phpunit/issues/3728)

michalbundyra:
Thanks, @Zombaya!

PR #34, opened 2020-01-14
Fix hostname validation for domain parts with 2+ dashes and with dash at the end

ingwer:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Without it, domains like `do-main-.com` will pass validation


michalbundyra:
Thanks, @ingwer!

PR #51, opened 2020-02-06
Fix some minors typos

codisart:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description
replace cannont by cannot
use only lowcase cannot
use only uppercase constants


michalbundyra:
Thanks, @codisart!

PR #55, opened 2020-02-14
Validators "Using a Select object" example is invalid

metalinspired:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Select->where is a Where predicate which, in turns, does not have such (magic) property


michalbundyra:
Thanks, @metalinspired!

PR #59, opened 2020-03-29
bugfix: allows Laminas\Uri\Uri as uriHandler in Uri validator

codisart:
|    Q          |   A
|-------------- | ------
| Bugfix | yes
| QA            | yes

### Description
Add constructor test for traversable object
Add constructor tests for invalid uri handler
Remode dead code on urihandler validation
Replace assertion calls by static calls
Replace `Laminas\Uri\Uri` by the class keyword 
Use the same alias `UriHandler` for `Laminas\Uri\Uri` as in the `Validator\Uri` class

### DIFF
--- a/src/Uri.php
+++ b/src/Uri.php
@@ -89,22 +89,17 @@ class Uri extends AbstractValidator
             // Instantiate string Uri handler that references a class
             $this->uriHandler = new $this->uriHandler;
         }
-
-        if (! $this->uriHandler instanceof UriHandler) {
-            throw new InvalidArgumentException('URI handler is expected to be a Laminas\Uri\Uri object');
-        }
-
         return $this->uriHandler;
     }
 
     /**
-     * @param  UriHandler $uriHandler
+     * @param  UriHandler|string  $uriHandler
      * @throws InvalidArgumentException
      * @return Uri
      */
     public function setUriHandler($uriHandler)
     {
-        if (! is_subclass_of($uriHandler, 'Laminas\Uri\Uri')) {
+        if (! is_a($uriHandler, UriHandler::class, true)) {
             throw new InvalidArgumentException(sprintf(
                 'Expecting a subclass name or instance of %s as $uriHandler',
                 UriHandler::class
--- a/test/UriTest.php
+++ b/test/UriTest.php
@@ -10,8 +10,9 @@ namespace LaminasTest\Validator;
 
 use Laminas\Uri\Exception\InvalidArgumentException;
 use Laminas\Uri\Http;
-use Laminas\Uri\Uri;
+use Laminas\Uri\Uri as UriHandler;
 use Laminas\Validator;
+use LaminasTest\Validator\TestAsset;
 use PHPUnit\Framework\TestCase;
 
 /**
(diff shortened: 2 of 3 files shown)


src/Uri.php
```diff
@@ -89,22 +89,17 @@ public function getUriHandler()
...
-        if (! $this->uriHandler instanceof UriHandler) {
-            throw new InvalidArgumentException('URI handler is expected to be a Laminas\Uri\Uri object');
-        }
-
         return $this->uriHandler;
     }
 
     /**
-     * @param  UriHandler $uriHandler
+     * @param  UriHandler | string  $uriHandler
```
michalbundyra: No space around `|` please.

src/Uri.php
```diff
@@ -89,22 +89,17 @@ public function getUriHandler()
...
      * @throws InvalidArgumentException
      * @return Uri
      */
     public function setUriHandler($uriHandler)
     {
-        if (! is_subclass_of($uriHandler, 'Laminas\Uri\Uri')) {
+        if (! is_subclass_of($uriHandler, UriHandler::class)) {
```
michalbundyra: I believe it is a bug in that place.
I think we should use `is_a` instead, as we should be allowing `UriHandler` instance as well. At least this is what is in PHPDocs and exception message. We would need, of course, a new test case.

test/UriTest.php
```diff
@@ -38,61 +38,75 @@ public function testHasDefaultSettingsAndLazyLoadsUriHandler()
...
-        $this->assertFalse($validator->getAllowAbsolute());
+        self::assertSame($uriMock, $validator->getUriHandler());
+        self::assertFalse($validator->getAllowRelative());
+        self::assertFalse($validator->getAllowAbsolute());
+    }
+
+    public function testConstructWithTraversableSetsOptions()
+    {
+        $uriMock = $this->createMock(UriHandler::class);
+        $options = new \ArrayObject([
```
michalbundyra: `ArrayObject` is very specific `Traversable` (it is also instance of `ArrayAccess` and we had cases when it was abused). That's why I prefer provide "a custom implementation" for Traversable, see what I've done in:
https://github.com/laminas/laminas-form/pull/55/files
and another PR where I was fixing usage of Traversable:
https://github.com/laminas/laminas-form/pull/60

It looks like the current implementation in `Uri` validator is valid, as we convert `Traversable` into array.

test/UriTest.php
```diff
@@ -38,61 +38,75 @@ public function testHasDefaultSettingsAndLazyLoadsUriHandler()
...
+            // allowAbsolute allowRelative isAbsolute isRelative isValid expects
+            [  true,         true,         true,      false,     true,   true],
+            [  true,         true,         false,     true,      true,   true],
+            [  false,        true,         true,      false,     true,   false],
+            [  false,        true,         false,     true,      true,   true],
+            [  true,         false,        true,      false,     true,   true],
+            [  true,         false,        false,     true,      true,   false],
+            [  false,        false,        true,      false,     true,   false],
+            [  false,        false,        false,     true,      true,   false],
+            [  true,         true,         false,     false,     false,  false],
```
michalbundyra: Please revert

test/UriTest.php
```diff
@@ -160,6 +174,18 @@ public function testUriHandlerInvalidTypeThrowsException()
         $this->validator->setUriHandler(new \stdClass());
     }
 
+    public function testConstructUriHandlerStringInvalidClassThrowsException()
+    {
+        $this->expectException(Validator\Exception\InvalidArgumentException::class);
```
michalbundyra: can we add here also expectExceptionMessage, so we know which argument is invalid

test/UriTest.php
```diff
@@ -160,6 +174,18 @@ public function testUriHandlerInvalidTypeThrowsException()
...
+        new Validator\Uri(\stdClass::class, false, false);
+    }
+
+    public function testConstructUriHandlerInvalidTypeThrowsException()
+    {
+        $this->expectException(Validator\Exception\InvalidArgumentException::class);
```
michalbundyra: as above please

src/Uri.php
codisart: I removed the space

src/Uri.php
codisart: I replaced by `is_a`  with the `allow_string` param as true.

test/UriTest.php
codisart: I added the `CustomTraversable` and used it instead of the `ArrayObject`

test/UriTest.php
codisart: reverted

test/UriTest.php
codisart: added

test/UriTest.php
codisart: added

froschdesign:
@codisart 
Thank you for your work!

For the next pull requests it would be great if you can split the all-in-one commit into smaller pieces. Separated commits allows to follow the commit history and to make cherry-picks.
Thanks in advance!

michalbundyra:
Thanks, @codisart!

PR #60, opened 2020-04-18
proposal: refactoring of options configuration

codisart:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

This pull request is my proposal to refactor the options system for the validator objects.
I have two propositions. 

The first one is to use a simple trait with a single public method to reduce the duplication of the code that transform function argumetns into an array of options 

The second one is to use a full fleged object to do the transformation. It is in my opinion heavier to use but I think it could be the base of more refactoring of the options system. Maybe adding validation.

### DIFF
--- a/src/Between.php
+++ b/src/Between.php
@@ -76,17 +76,8 @@ class Between extends AbstractValidator
             $options = ArrayUtils::iteratorToArray($options);
         }
         if (! is_array($options)) {
-            $options = func_get_args();
-            $temp['min'] = array_shift($options);
-            if (! empty($options)) {
-                $temp['max'] = array_shift($options);
-            }
-
-            if (! empty($options)) {
-                $temp['inclusive'] = array_shift($options);
-            }
-
-            $options = $temp;
+            $validatorOptions = new ValidatorOptionsObject(['min', 'max', 'inclusive']);
+            $options = $validatorOptions->argumentsAsArray(func_get_args());
         }
 
         if (! array_key_exists('min', $options) || ! array_key_exists('max', $options)) {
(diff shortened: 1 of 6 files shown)


glensc:
perhaps drop the trait, but rather add static methods to ValidatorOptionsObject for factory patterns.

codisart:
Hello @glensc, thank you for the review.

I like your idea, what do you think of 
```php
$options = ValidatorOptions::asArray(['min', 'max', 'inclusive'], func_get_args());
```
```php
class ValidatorOptions
{
    public static asArray(array $keys, array $args) {
        $this->shouldHaveOnlyStringValues($keys);

        $result = [];
        foreach ($args as $arg) {
            $result[array_shift($keys)] = $arg;
        }
        return $result;
    }

    private function shouldHaveOnlyStringValues(array $keys) : void
    {
        foreach ($keys as $key) {
            if (! is_string($key)) {
                throw new \InvalidArgumentException('The values of $keys should be only strings');
            }
        }
    }
}
```

boesing:
@codisart Could you please create an RFC issue instead of a PR? 
I think we might want to have some more feedback on what we want to change and having an issue without a concrete implementation detail could engage more people in participating the discussion.

codisart:
Hello @boesing,

I never did a RFC issue. Do you have an exemple to follow ?

boesing:
@codisart Sure, have a look at [discourse](https://discourse.laminas.dev/t/rfc-opcache-preloading-for-mezzio-and-mvc/1442) for now, as we are trying to move RFCs to github issues since the november TSC meeting. 👍

gsteel:
Thanks for this contribution @codisart but it's got so old, I'm going to close it. For [3.0](https://github.com/laminas/laminas-validator/milestone/5?closed=1) we've stuck with regular arrays annotated with psalm types.

PR #61, opened 2020-04-28
Fix invalid option name in callback validator

metalinspired:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description
Additional options option should be named 'callbackOptions' instead of 'options'

### DIFF
--- a/docs/book/validators/callback.md
+++ b/docs/book/validators/callback.md
@@ -8,7 +8,7 @@ validate a given value.
 The following options are supported for `Laminas\Validator\Callback`:
 
 - `callback`: Sets the callback which will be called for the validation.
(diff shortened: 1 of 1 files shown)


samsonasik:
DCO is required, please check the detail https://github.com/laminas/laminas-validator/pull/61/checks?check_run_id=625546112

michalbundyra:
Thanks, @metalinspired!

michalbundyra:
@samsonasik DCO is not needed in that case, see:
https://github.com/laminas/.github/blob/master/CONTRIBUTING.md#obvious-fix-policy

PR #62, opened 2020-05-10
qa: Replace array by data providers in unit tests

codisart:

|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Replace array for configuration of tests by data providers. 
It allows a better isolation of failing tests

### DIFF
--- a/test/DigitsTest.php
+++ b/test/DigitsTest.php
@@ -16,9 +16,7 @@ use PHPUnit\Framework\TestCase;
  */
 class DigitsTest extends TestCase
 {
-    /**
-     * @var Digits
-     */
+    /** @var Digits */
     protected $validator;
 
     protected function setUp() : void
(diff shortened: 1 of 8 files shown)


test/DigitsTest.php
```diff
@@ -29,24 +29,27 @@ protected function setUp() : void
...
-        foreach ($valuesExpected as $input => $result) {
-            $this->assertEquals($result, $this->validator->isValid($input));
-        }
+        $this->assertSame($expected, $this->validator->isValid($input));
+    }
+
+    public function basicDataProvider()
+    {
+        return [
+            ['abc123',  false],
```
weierophinney: Can I ask for one further improvement, please? Add _names_ to each of these, to describe the specific behavior:

```php
'invalid; starts with alphabetic chars' => ['abc123', false],
```

and so on.

This helps us identify _specific_ regressions later - and makes it easier to find the test case (and thus the data being tested) when there are failures.

test/File/FilesSizeTest.php
```diff
@@ -28,50 +28,55 @@ protected function setUp() : void
...
+        );
+    }
 
-        $validator = new File\FilesSize(['min' => 0, 'max' => 200]);
-        $this->assertEquals(false, $validator->isValid(__DIR__ . '/_files/nofile.mo'));
-        $this->assertArrayHasKey('fileFilesSizeNotReadable', $validator->getMessages());
+    public function basicDataProvider()
+    {
+        return [
+            [   [['min' => 0, 'max' => 2000]],      true,   true,   false],
```
weierophinney: With this, and other datasets where you add whitespace around arrays for readability, surround the cases with:

```php
// phpcs:disable
/* add all the various cases */
// phpcs:enable
```

Otherwise, we'll get new phpcs failures, while fixing tests.

test/HexTest.php
```diff
@@ -29,11 +29,17 @@ protected function setUp() : void
     /**
      * Ensures that the validator follows expected behavior
      *
+     * @dataProvider basicDataProvider
      * @return void
      */
-    public function testBasic()
+    public function testBasic($input, $expected)
```
weierophinney: For any test method that is changed to use a data provider, please add `@param` annotations for any non-typehinted arguments.

weierophinney (changes requested):
This is fantastic, and something I've been planning to do for years, but never been able to prioritize!

I've made some notes for additional improvements, which will make the changes you're introducing even more solid. Looking forward to your final results!

test/HexTest.php
codisart: Should I keep the `@return` ?

codisart: Do the type hinting suffice if it is possible ?

test/DigitsTest.php
codisart: I added for every cases. What do you think about it ?

test/File/FilesSizeTest.php
codisart: I added the comments

weierophinney (approved):

PR #63, opened 2020-05-10
qa: Replace array by data providers in unit tests for the hostname validator

codisart:

|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Replace array for configuration of tests by data providers.
It allows a better isolation of failing tests

### DIFF
--- a/test/HostnameTest.php
+++ b/test/HostnameTest.php
@@ -45,112 +45,96 @@ class HostnameTest extends TestCase
     /**
      * Ensures that the validator follows expected behavior
      *
+     * @dataProvider basicDataProvider
+     * @param int $option
+     * @param bool $expected
+     * @param string $hostname
      * @return void
      */
-    public function testBasic()
+    public function testBasic(int $option, bool $expected, string $hostname)
     {
-        $valuesExpected = [
-            [Hostname::ALLOW_IP, true, ['1.2.3.4', '10.0.0.1', '255.255.255.255']],
-            [Hostname::ALLOW_IP, false, ['1.2.3.4.5', '0.0.0.256']],
-            [Hostname::ALLOW_DNS, true, ['example.com', 'example.museum', 'd.hatena.ne.jp', 'example.photography']],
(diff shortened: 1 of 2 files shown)


test/HostnameTest.php
```diff
@@ -45,11 +45,20 @@ protected function tearDown() : void
     /**
      * Ensures that the validator follows expected behavior
      *
+     * @dataProvider basicDataProvider
      * @return void
      */
-    public function testBasic()
+    public function testBasic($option, $expected, $inputValues)
```
weierophinney: Please add `@param` annotations for these new arguments.

test/HostnameTest.php
```diff
@@ -45,11 +45,20 @@ protected function tearDown() : void
...
+        $validator = new Hostname($option);
+        foreach ($inputValues as $input) {
+            $this->assertSame($expected, $validator->isValid($input));
+        }
+    }
+
+    public function basicDataProvider()
+    {
+        return [
             [Hostname::ALLOW_IP, true, ['1.2.3.4', '10.0.0.1', '255.255.255.255']],
```
weierophinney: I'd argue that you should split these out, so we can have failures rise for each of the IP/hostname arguments; that will allow us to identify specific recursions later more easily.

Additionally, provide names for each case:

```php
'allow-ip succeeds for 1.2.3.4' => [Hostname::ALLOW_IP, true, '1.2.3.4'],
```

As noted in a PR you've made against laminas-validator, this helps us identify specific data sets more easily when they fail.

weierophinney (changes requested):
Thanks for these improvements! I've add a couple notes for further improvement; while they're at specific locations, they apply generally to your entire patch.

test/HostnameTest.php
codisart: I have added the annotations

test/HostnameTest.php
codisart: I have added the keys and separated the test cases

PR #67, opened 2020-08-14
Possible fixes #8 compilation failed regex too large

### LINKED ISSUE #8: preg_match(): Compilation failed: regular expression is too large at offset 156184
This happens while using `Zend\Validator\Hostname->isValid('tapi4457@hsoqvf.biz')` because hostname/Biz.php contains enormous regular expression

Trace looks like this:

preg_match('/^[\\x{002d}0-9a-z\\x{3447}\\x{3473}\\x{359..(156088 chars skipped)..9FA0}\\x{9FA2}\\x{9FA4}\\x{9FA5}]{1,20}$/iu','tapi4457@hsoqvf')
  /vendor/zendframework/zend-validator/src/Hostname.php:1877

PHP Version 7.1.7

---

Originally posted by @ac-dc at https://github.com/zendframework/zend-validator/issues/219

samsonasik:
Signed-off-by: Abdul Malik Ikhsan <samsonasik@gmail.com>

|    Q          |   A
|-------------- | ------
| Bugfix        | yes

### Description

This is a possible quick fix for #8 . I am not sure if it is correct, but hostname with `@` seems invalid.

### DIFF
--- a/src/Hostname.php
+++ b/src/Hostname.php
@@ -1977,6 +1977,9 @@ class Hostname extends AbstractValidator
             return true;
         }
 
+        // Handle Regex compilation failure that may happen on .biz domain with has @ character, eg: tapi4457@hsoqvf.biz
+        // Technically, hostname with '@' character is invalid, so mark as invalid immediately
+        // @see https://github.com/laminas/laminas-validator/issues/8
         if (strpos($value, '@') !== false) {
             $this->error(self::INVALID_HOSTNAME);
             return false;
--- a/src/Hostname.php
+++ b/src/Hostname.php
@@ -1977,6 +1977,11 @@ class Hostname extends AbstractValidator
             return true;
         }
 
+        if (mb_strpos($value, '@') !== false) {
+            $this->error(self::INVALID_HOSTNAME);
+            return false;
+        }
+
         // Local hostnames are allowed to be partial (ending '.')
         if ($this->getAllow() & self::ALLOW_LOCAL) {
             if (substr($value, -1) === '.') {
--- a/src/Hostname.php
+++ b/src/Hostname.php
@@ -1977,7 +1977,7 @@ class Hostname extends AbstractValidator
             return true;
         }
 
-        if (mb_strpos($value, '@') !== false) {
+        if (strpos($value, '@') !== false) {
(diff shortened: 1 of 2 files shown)


samsonasik:
@ac-dc @Thaix could you verify if this patch works for you ?

Thaix:
Yes, it works on my machine!
Before:

```
$ php -a
Interactive mode enabled

php > include 'vendor/autoload.php';
php > var_dump((new Laminas\Validator\Hostname())->isValid('tapi4457@hsoqvf.biz'));
PHP Warning:  preg_match(): Compilation failed: regular expression is too large at offset 156184 in /home/apidev/current/vendor/laminas/laminas-validator/src/Hostname.php on line 2104
bool(false)
```

After
```
php -a
Interactive mode enabled

php > include 'vendor/autoload.php';
php > var_dump((new Laminas\Validator\Hostname())->isValid('tapi4457@hsoqvf.biz'));
bool(false)
```

src/Hostname.php
```diff
@@ -1977,6 +1977,11 @@ public function isValid($value)
             return true;
         }
 
+        if (strpos($value, '@') !== false) {
```
weierophinney, line 1983: Please add a comment above this one indicating:

- The issue it resolves
- What the root cause is that it's resolving (`@` in a hostname is always invalid, so there's no need to use a regex to detect that scenario).

weierophinney (commented):
This technically solves the reported issue. I'm a bit worried that we'll still see issues crop up around the "regexp compilation failed" due to regexp size, though. I wonder if it might be good to try and break it into a number of smaller `strpos()` checks for common, known-bad characters, which would allow us to reduce what the regexp looks for?

src/Hostname.php
samsonasik: I've added the comment.

samsonasik:
@weierophinney do we have list of known-bad characters for hostname?

weierophinney:
> do we have list of known-bad characters for hostname?

I think part of the complexity of the regex is a list of characters to exclude, IIRC. :smile:

samsonasik:
How about merge for now?

samsonasik:
@weierophinney I think we can merge as is for now, as concrete solution seems not imaginable right now.

PR #72, opened 2020-09-30
Fixed parameter types

svycka:


src/AbstractValidator.php
```diff
@@ -321,8 +321,8 @@ protected function createMessage($messageKey, $value)
     }
 
     /**
-     * @param  string $messageKey
-     * @param  string $value      OPTIONAL
+     * @param  string              $messageKey
+     * @param  string|array|object $value      OPTIONAL
```
michalbundyra: Also `null` should be noted here, imo.

svycka: then what createMessage() will do as it does not accept null :)

michalbundyra: It's fine, see implementation:

```php
if ($value === null) {
    $value = $this->value;
}
```

svycka:
but default value is still null

boesing:
@svycka Thanks for this PR.

To make travis pass, you can either temporarily add `COMPOSER_ROOT_VERSION=2.14.0` to `travis.yml` or `"version": "2.14.0"` to `composer.json` in a single commit.
After this, travis should startup and execute tests.
When anything is green, we can add this to the next release 👍

svycka:
@boesing I am A bit lost do you want something changed? I got an email but I guess you removed the comment.

boesing:
Nah, everything is fine. Thats why I deleted my comment.
Just have to sort some things out regarding this project and merging stuff as we have commits in here which should already be part of the branch but isn't...

PR #75, opened 2020-10-05
PHP 8.0 support

### LINKED ISSUE #74: PHP 8.0 support
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes

#### Summary

To be prepared for the december release of PHP 8.0, this repository has some additional TODOs to be tested against the new major version.

In order to make this repository compatible, one has to follow these steps:

- [ ] Modify `composer.json` to provide support for PHP 8.0 by adding the constraint `~8.0.0`
- [ ] Modify `composer.json` to drop support for PHP less than 7.3
- [ ] Modify `composer.json` to implement phpunit 9.3 which supports PHP 7.3+
- [ ] Modify `.travis.yml` to ignore platform requirements when installing composer dependencies (simply add `--ignore-platform-reqs` to `COMPOSER_ARGS` env variable)
- [ ] Modify `.travis.yml` to add PHP 8.0 to the matrix (*NOTE*: Do not allow failures as PHP 8.0 has a feature freeze since 2020-08-04!)
- [ ] Modify source code in case there are incompatibilities with PHP 8.0

jfernancordova:
**_Issue_** #74 

|    Q          |   A
|-------------- | ------
| New Feature   | yes

### Description

- [x] Modify composer.json to provide support for PHP 8.0 by adding the constraint ~8.0.0
- [x]  Modify composer.json to drop support for PHP less than 7.3
- [x] Modify composer.json to implement phpunit 9.3 which supports PHP 7.3+
- [x] Modify .travis.yml to ignore platform requirements when installing composer dependencies (simply add --ignore-platform-reqs to COMPOSER_ARGS env variable)
- [x] Modify .travis.yml to add PHP 8.0 to the matrix (NOTE: Do not allow failures as PHP 8.0 has a feature freeze since 2020-08-04!)
- [x] Modify source code in case there are incompatibilities with PHP 8.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,33 +13,29 @@ env:
 matrix:
   fast_finish: true
   include:
-    - php: 7.1
+    - php: 7.3
       env:
         - DEPS=lowest
-    - php: 7.1
+    - php: 7.3
+      env:
+        - DEPS=latest
+    - php: 7.4
+      env:
+        - DEPS=lowest
+    - php: 7.4
       env:
         - DEPS=latest
         - EXECUTE_HOSTNAME_CHECK=true
         - CS_CHECK=true
         - TEST_COVERAGE=true
-    - php: 7.2
-      env:
-        - DEPS=lowest
-    - php: 7.2
+    - php: 8.0
       env:
         - DEPS=latest
-    - php: 7.3
+        - COMPOSER_ARGS="--ignore-platform-reqs"
+    - php: 8.0
       env:
         - DEPS=lowest
-    - php: 7.3
-      env:
-        - DEPS=latest
-    - php: 7.4
-      env:
-        - DEPS=lowest
-    - php: 7.4
-      env:
-        - DEPS=latest
+        - COMPOSER_ARGS="--ignore-platform-reqs"
 
 before_install:
   - if [[ $TEST_COVERAGE != 'true' ]]; then phpenv config-rm xdebug.ini || return 0 ; fi
--- a/composer.json
+++ b/composer.json
@@ -25,9 +25,9 @@
         }
     },
     "require": {
-        "php": "^7.1",
+        "php": "^7.3 || ~8.0.0",
         "container-interop/container-interop": "^1.1",
-        "laminas/laminas-stdlib": "^3.2.1",
+        "laminas/laminas-stdlib": "^3.3",
         "laminas/laminas-zendframework-bridge": "^1.0"
     },
     "require-dev": {
@@ -36,13 +36,14 @@
         "laminas/laminas-config": "^2.6",
         "laminas/laminas-db": "^2.7",
         "laminas/laminas-filter": "^2.6",
-        "laminas/laminas-http": "^2.5.4",
+        "laminas/laminas-http": "^2.14.2",
         "laminas/laminas-i18n": "^2.6",
         "laminas/laminas-math": "^2.6",
         "laminas/laminas-servicemanager": "^2.7.5 || ^3.0.3",
         "laminas/laminas-session": "^2.8",
-        "laminas/laminas-uri": "^2.5",
-        "phpunit/phpunit": "^7.5.20 || ^8.5.2",
+        "laminas/laminas-uri": "^2.7",
+        "phpspec/prophecy-phpunit": "^2.0",
+        "phpunit/phpunit": "^9.3",
         "psr/http-client": "^1.0",
         "psr/http-factory": "^1.0",
         "psr/http-message": "^1.0"
--- a/src/InArray.php
+++ b/src/InArray.php
@@ -197,10 +197,6 @@ class InArray extends AbstractValidator
                     if ($el == $value) {
                         return true;
                     }
-
-                    if (self::COMPARE_NOT_STRICT == $this->strict) {
-                        return true;
-                    }
                 }
             }
         } else {
--- a/src/InArray.php
+++ b/src/InArray.php
@@ -197,6 +197,10 @@ class InArray extends AbstractValidator
                     if ($el == $value) {
                         return true;
                     }
+
+                    if ($this->isMultidimensionalArray($haystack) && self::COMPARE_NOT_STRICT == $this->strict) {
+                        return true;
+                    }
                 }
             }
         } else {
@@ -216,14 +220,32 @@ class InArray extends AbstractValidator
                         $h = (string) $h;
                     }
                 }
+
+                if (! $this->isMultidimensionalArray($haystack) && in_array($value, $haystack, $this->strict)) {
+                    return true;
+                }
             }
 
-            if (in_array($value, $haystack, self::COMPARE_STRICT == $this->strict)) {
+            if (array_search($value, $haystack, self::COMPARE_STRICT == $this->strict)) {
+                return true;
+            }
+
+            if (self::COMPARE_NOT_STRICT == $this->strict) {
                 return true;
             }
         }
 
         $this->error(self::NOT_IN_ARRAY);
+        return false;
+    }
+
+    private function isMultidimensionalArray(array $array): bool
+    {
+        foreach ($array as $value) {
+            if (is_array($value)) {
+                return true;
+            }
+        }
         return false;
     }
 }
--- a/src/InArray.php
+++ b/src/InArray.php
@@ -198,7 +198,7 @@ class InArray extends AbstractValidator
                         return true;
                     }
 
-                    if ($this->isMultidimensionalArray($haystack) && self::COMPARE_NOT_STRICT == $this->strict) {
+                    if (self::COMPARE_NOT_STRICT == $this->strict) {
                         return true;
                     }
                 }
@@ -221,7 +221,7 @@ class InArray extends AbstractValidator
                     }
                 }
 
-                if (! $this->isMultidimensionalArray($haystack) && in_array($value, $haystack, $this->strict)) {
+                if (in_array($value, $haystack, $this->strict)) {
                     return true;
                 }
             }
@@ -236,16 +236,6 @@ class InArray extends AbstractValidator
         }
 
         $this->error(self::NOT_IN_ARRAY);
-        return false;
-    }
-
-    private function isMultidimensionalArray(array $array): bool
-    {
-        foreach ($array as $value) {
-            if (is_array($value)) {
-                return true;
-            }
-        }
         return false;
     }
 }
--- a/src/InArray.php
+++ b/src/InArray.php
@@ -185,50 +185,58 @@ class InArray extends AbstractValidator
                     if ($element === $value) {
                         return true;
                     }
-                } else {
-                    // add protection to prevent string to int vuln's
-                    $el = $element;
-                    if (self::COMPARE_NOT_STRICT_AND_PREVENT_STR_TO_INT_VULNERABILITY == $this->strict
-                        && is_string($value) && (is_int($el) || is_float($el))
-                    ) {
-                        $el = (string) $el;
-                    }
 
-                    if ($el == $value) {
-                        return true;
-                    }
-                }
-            }
-        } else {
-            /**
-             * If the check is not strict, then, to prevent "asdf" being converted to 0
-             * and returning a false positive if 0 is in haystack, we type cast
-             * the haystack to strings. To prevent "56asdf" == 56 === TRUE we also
-             * type cast values like 56 to strings as well.
-             *
-             * This occurs only if the input is a string and a haystack member is an int
-             */
-            if (self::COMPARE_NOT_STRICT_AND_PREVENT_STR_TO_INT_VULNERABILITY == $this->strict
-                && is_string($value)
-            ) {
-                foreach ($haystack as &$h) {
-                    if (is_int($h) || is_float($h)) {
-                        $h = (string) $h;
-                    }
+                    continue;
                 }
 
-                if (in_array($value, $haystack, $this->strict)) {
+                // add protection to prevent string to int vuln's
+                $el = $element;
+                if (self::COMPARE_NOT_STRICT_AND_PREVENT_STR_TO_INT_VULNERABILITY == $this->strict
+                    && is_string($value) && (is_int($el) || is_float($el))
+                ) {
+                    $el = (string) $el;
+                }
+
+                if ($el == $value) {
                     return true;
                 }
             }
 
-            if (array_search($value, $haystack, self::COMPARE_STRICT == $this->strict)) {
+            $this->error(self::NOT_IN_ARRAY);
+            return false;
+        }
+
+        /**
+         * If the check is not strict, then, to prevent "asdf" being converted to 0
+         * and returning a false positive if 0 is in haystack, we type cast
+         * the haystack to strings. To prevent "56asdf" == 56 === TRUE we also
+         * type cast values like 56 to strings as well.
+         *
+         * This occurs only if the input is a string and a haystack member is an int
+         */
+        if (self::COMPARE_NOT_STRICT_AND_PREVENT_STR_TO_INT_VULNERABILITY == $this->strict
+            && is_string($value)
+        ) {
+            foreach ($haystack as &$h) {
+                if (is_int($h) || is_float($h)) {
+                    $h = (string) $h;
+                }
+            }
+
+            if (in_array($value, $haystack, (bool) $this->strict)) {
                 return true;
             }
 
-            if (self::COMPARE_NOT_STRICT == $this->strict) {
-                return true;
-            }
+            $this->error(self::NOT_IN_ARRAY);
+            return false;
+        }
+
+        if (in_array($value, $haystack, self::COMPARE_STRICT == $this->strict)) {
+            return true;
+        }
+
+        if (self::COMPARE_NOT_STRICT == $this->strict) {
+            return true;
         }
 
         $this->error(self::NOT_IN_ARRAY);
--- a/test/Db/NoRecordExistsTest.php
+++ b/test/Db/NoRecordExistsTest.php
@@ -228,10 +228,7 @@ class NoRecordExistsTest extends TestCase
     public function testEqualsMessageTemplates()
     {
         $validator  = new NoRecordExists('users', 'field1');
-        $this->assertAttributeEquals(
-            $validator->getOption('messageTemplates'),
-            'messageTemplates',
-            $validator
-        );
+        $this->assertObjectHasAttribute('messageTemplates', $validator);
+        $this->assertEquals($validator->getOption('messageTemplates'), $validator->getMessageTemplates());
     }
 }
--- a/test/InArrayTest.php
+++ b/test/InArrayTest.php
@@ -230,12 +230,14 @@ class InArrayTest extends TestCase
         $validator->setStrict(InArray::COMPARE_NOT_STRICT);
         $validator->setRecursive(true);
 
+        $stringToNumericComparisonAssertion = PHP_MAJOR_VERSION < 8 ? 'assertTrue' : 'assertFalse';
+
         $this->assertEquals(InArray::COMPARE_NOT_STRICT, $validator->getStrict());
-        $this->assertTrue($validator->isValid('b'));
+        $this->$stringToNumericComparisonAssertion($validator->isValid('b'));
         $this->assertTrue($validator->isValid('a'));
         $this->assertTrue($validator->isValid('A'));
         $this->assertTrue($validator->isValid('0'));
-        $this->assertTrue($validator->isValid('1a'));
+        $this->$stringToNumericComparisonAssertion($validator->isValid('1a'));
         $this->assertTrue($validator->isValid(0));
     }
 
--- a/test/InArrayTest.php
+++ b/test/InArrayTest.php
@@ -140,12 +140,13 @@ class InArrayTest extends TestCase
         $validator->setStrict(InArray::COMPARE_STRICT);
 
         $this->assertTrue($validator->getStrict());
+
+        $this->assertTrue($validator->isValid('A'));
+        $this->assertTrue($validator->isValid(0));
         $this->assertFalse($validator->isValid('b'));
         $this->assertFalse($validator->isValid('a'));
-        $this->assertTrue($validator->isValid('A'));
         $this->assertFalse($validator->isValid('0'));
         $this->assertFalse($validator->isValid('1a'));
-        $this->assertTrue($validator->isValid(0));
     }
 
     public function testNonStrictComparisons()
@@ -360,10 +361,8 @@ class InArrayTest extends TestCase
 
     public function testEqualsMessageTemplates()
     {
-        $this->assertAttributeEquals(
-            $this->validator->getOption('messageTemplates'),
-            'messageTemplates',
-            $this->validator
-        );
+        $validator = $this->validator;
+        $this->assertObjectHasAttribute('messageTemplates', $validator);
+        $this->assertEquals($validator->getOption('messageTemplates'), $validator->getMessageTemplates());
     }
 }
--- a/test/ValidatorPluginManagerFactoryTest.php
+++ b/test/ValidatorPluginManagerFactoryTest.php
@@ -15,9 +15,11 @@ use Laminas\Validator\ValidatorInterface;
 use Laminas\Validator\ValidatorPluginManager;
 use Laminas\Validator\ValidatorPluginManagerFactory;
 use PHPUnit\Framework\TestCase;
+use Prophecy\PhpUnit\ProphecyTrait;
 
 class ValidatorPluginManagerFactoryTest extends TestCase
 {
+    use ProphecyTrait;
     public function testFactoryReturnsPluginManager()
     {
         $container = $this->prophesize(ContainerInterface::class)->reveal();
@@ -25,14 +27,6 @@ class ValidatorPluginManagerFactoryTest extends TestCase
 
         $validators = $factory($container, ValidatorPluginManagerFactory::class);
         $this->assertInstanceOf(ValidatorPluginManager::class, $validators);
-
-        if (method_exists($validators, 'configure')) {
-            // laminas-servicemanager v3
-            $this->assertAttributeSame($container, 'creationContext', $validators);
-        } else {
-            // laminas-servicemanager v2
-            $this->assertSame($container, $validators->getServiceLocator());
-        }
     }
 
     /**
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -17,14 +17,15 @@
         </exclude>
     </groups>
 
-    <filter>
-        <whitelist addUncoveredFilesFromWhitelist="true">
+    <coverage processUncoveredFiles="true">
+        <include>
             <directory suffix=".php">./src</directory>
-        </whitelist>
-    </filter>
+        </include>
+    </coverage>
 
     <php>
         <ini name="date.timezone" value="UTC"/>
+        <ini name="xdebug.max_nesting_level" value="3000"/>
 
         <!-- OB_ENABLED should be enabled for some tests to check if all
              functionality works as expected. Such tests include those for
--- a/test/BarcodeTest.php
+++ b/test/BarcodeTest.php
@@ -487,20 +487,14 @@ class BarcodeTest extends TestCase
     public function testEqualsMessageTemplates()
     {
         $validator = new Barcode('code25');
-        $this->assertAttributeEquals(
-            $validator->getOption('messageTemplates'),
-            'messageTemplates',
-            $validator
-        );
+        $this->assertObjectHasAttribute('messageTemplates', $validator);
+        $this->assertEquals($validator->getOption('messageTemplates'), $validator->getMessageTemplates());
     }
 
     public function testEqualsMessageVariables()
     {
         $validator = new Barcode('code25');
-        $this->assertAttributeEquals(
-            $validator->getOption('messageVariables'),
-            'messageVariables',
-            $validator
-        );
+        $this->assertObjectHasAttribute('messageVariables', $validator);
+        $this->assertEquals(array_keys($validator->getOption('messageVariables')), $validator->getMessageVariables());
     }
 }
--- a/test/BetweenTest.php
+++ b/test/BetweenTest.php
@@ -238,13 +238,15 @@ class BetweenTest extends TestCase
     public function testEqualsMessageTemplates()
     {
         $validator = new Between(['min' => 1, 'max' => 10]);
-        $this->assertAttributeEquals($validator->getOption('messageTemplates'), 'messageTemplates', $validator);
+        $this->assertObjectHasAttribute('messageTemplates', $validator);
+        $this->assertEquals($validator->getOption('messageTemplates'), $validator->getMessageTemplates());
     }
 
     public function testEqualsMessageVariables()
     {
         $validator = new Between(['min' => 1, 'max' => 10]);
-        $this->assertAttributeEquals($validator->getOption('messageVariables'), 'messageVariables', $validator);
+        $this->assertObjectHasAttribute('messageVariables', $validator);
+        $this->assertEquals(array_keys($validator->getOption('messageVariables')), $validator->getMessageVariables());
     }
 
     /**
--- a/test/CallbackTest.php
+++ b/test/CallbackTest.php
@@ -77,11 +77,8 @@ class CallbackTest extends TestCase
     public function testEqualsMessageTemplates()
     {
         $validator = new Callback([$this, 'objectCallback']);
-        $this->assertAttributeEquals(
-            $validator->getOption('messageTemplates'),
-            'messageTemplates',
-            $validator
-        );
+        $this->assertObjectHasAttribute('messageTemplates', $validator);
+        $this->assertEquals($validator->getOption('messageTemplates'), $validator->getMessageTemplates());
     }
 
     public function testCanAcceptContextWithoutOptions()
--- a/test/CreditCardTest.php
+++ b/test/CreditCardTest.php
@@ -362,11 +362,8 @@ class CreditCardTest extends TestCase
     public function testEqualsMessageTemplates()
     {
         $validator = new CreditCard();
-        $this->assertAttributeEquals(
-            $validator->getOption('messageTemplates'),
-            'messageTemplates',
-            $validator
-        );
+        $this->assertObjectHasAttribute('messageTemplates', $validator);
+        $this->assertEquals($validator->getOption('messageTemplates'), $validator->getMessageTemplates());
     }
 
     /**
--- a/test/DateStepTest.php
+++ b/test/DateStepTest.php
@@ -110,7 +110,7 @@ class DateStepTest extends TestCase
     {
         $validator  = new Validator\DateStep([]);
         $this->assertObjectHasAttribute('messageTemplates', $validator);
-        $this->assertAttributeEquals($validator->getOption('messageTemplates'), 'messageTemplates', $validator);
+        $this->assertEquals($validator->getOption('messageTemplates'), $validator->getMessageTemplates());
     }
 
     public function testStepError()
--- a/test/DateTest.php
+++ b/test/DateTest.php
@@ -160,21 +160,15 @@ class DateTest extends TestCase
     public function testEqualsMessageTemplates()
     {
         $validator = $this->validator;
-        $this->assertAttributeEquals(
-            $validator->getOption('messageTemplates'),
-            'messageTemplates',
-            $validator
-        );
+        $this->assertObjectHasAttribute('messageTemplates', $validator);
+        $this->assertEquals($validator->getOption('messageTemplates'), $validator->getMessageTemplates());
     }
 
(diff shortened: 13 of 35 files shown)


composer.json
```diff
@@ -25,7 +25,7 @@
         }
     },
     "require": {
-        "php": "^7.1",
+        "php": "~8.0.0",
```
boesing: Please use `^7.3 || ~8.0.0` as descripted in #74

composer.json
```diff
@@ -42,7 +42,7 @@
         "laminas/laminas-servicemanager": "^2.7.5 || ^3.0.3",
         "laminas/laminas-session": "^2.8",
         "laminas/laminas-uri": "^2.5",
-        "phpunit/phpunit": "^7.5.20 || ^8.5.2",
+        "phpunit/phpunit": "^9.3.*",
```
boesing: Its okay to use `^9.3` here

.travis.yml
```diff
@@ -6,7 +6,7 @@ cache:
 
 env:
   global:
-    - COMPOSER_ARGS="--no-interaction"
+    - COMPOSER_ARGS="--ignore-platform-reqs --no-interaction"
```
boesing: please only ignore platform requirements for nightly (which is not yet part of this file)

boesing (changes requested):

composer.json
jfernancordova: Fixed

composer.json
jfernancordova: Fixed

.travis.yml
jfernancordova: Ok, I get it, I'll revert this

jfernancordova:
@boesing I just added PHP 8.0 to the matrix in travis.yml, anything else?

boesing:
> @boesing I just added PHP 8.0 to the matrix in travis.yml, anything else?

You have to use `nightly` in travis as there is no `8.0` version available yet.
After this, you can convert your PR to `Ready for review` by yourself and then we'll see what travis says 👍

jfernancordova:
Perfect

boesing:
@jfernancordova Please check mezzio/mezzio-cors [.travis.yml](https://github.com/mezzio/mezzio-cors/blob/1.1.x/.travis.yml) to see how you have to change the `.travis.yml` to finish this PR. Just ignore the `allow_failures` part as this is fixed in a PR in that repository.

If travis is "green", this is done 🎉

boesing (changes requested):
Please remove PHP <7.3 from travis aswell.

jfernancordova:
@boesing it's failing for **_laminas/laminas-uri_**, do you think that upgrading this dependency, could it solve this error in travis?

glensc:
`--ignore-platform-reqs` should never be used. it does not just ignore `ext-foo` requirement, but also PHP version dependency installing packages that are not compatible at all with the runtime. for example running in 5.6 installing package resolving that requires PHP 7.1 and then execution results fatal errors due invalid syntax. 

if you don't have ext-foo installed, fake it with local `composer config platform.ext-foo '0'` instead.

example:
- https://github.com/eventum/eventum/blob/v3.9.0/bin/releng/configure.sh#L11-L13

boesing:
> `--ignore-platform-reqs` should never be used. it does not just ignore `ext-foo` requirement, but also PHP version dependency installing packages that are not compatible at all with the runtime. for example running in 5.6 installing package resolving that requires PHP 7.1 and then execution results fatal errors due invalid syntax.
> 
> if you don't have ext-foo installed, fake it with local `composer config platform.ext-foo '0'` instead.
 
Thanks for the information.
You are right, its not a good solution. However, this is the only way to initially provide PHP 8.0 support on packages. Since we are not using `composer v2` which wont happen in the near future anyways, there is no way to use `--ignore-platform-req=php` and thus, we *have* to use `--ignore-platform-reqs`. Without ignoring platform reqs, this package is not testable by travis.
However, this should not be used in upstream projects but for travis this is absolutely fine.

.travis.yml
```diff
@@ -6,28 +6,13 @@ cache:
 
 env:
   global:
-    - COMPOSER_ARGS="--no-interaction"
+    - COMPOSER_ARGS="--no-interaction --ignore-platform-reqs"
```
boesing: Please only ignore platform reqs for nightly builds

boesing (changes requested):

.travis.yml
jfernancordova: Fixed

glensc:
> we have to use `--ignore-platform-reqs`. 

that is incorrect. what is the error you trying to avoid with the flag?

boesing:
> > we have to use `--ignore-platform-reqs`.
> 
> that is incorrect. what is the error you trying to avoid with the flag?

I want to avoid having trouble with 3rd-party composer dependencies which are not yet PHP 8.0 compatible as no dependency wont install when one of them are not compatible (yet).

Lets take `laminas/laminas-uri` for example. You wont be able to use `composer install` unless you either provide `--ignore-platform-reqs` (to ignore that the package is not yet compatible with PHP 8.0) or by using `--no-dev`.

Feel free to resolve dependency tree on PHP 8.0 without passing `--ignore-platform-reqs`.

But as you wanted to have the error:

```
PATH="/usr/local/Cellar/php@8.0/8.0.0/bin:$PATH" php80 /usr/local/bin/composer install
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for laminas/laminas-config ^2.6 -> satisfiable by laminas/laminas-config[2.6.0].
    - laminas/laminas-config 2.6.0 requires php ^5.5 || ^7.0 -> your PHP version (8.0.0-dev) does not satisfy that requirement.
  Problem 2
    - laminas/laminas-cache 2.9.0 requires php ^5.6 || ^7.0 -> your PHP version (8.0.0-dev) does not satisfy that requirement.
    - laminas/laminas-cache 2.8.3 requires php ^5.6 || ^7.0 -> your PHP version (8.0.0-dev) does not satisfy that requirement.
    - laminas/laminas-cache 2.8.2 requires php ^5.6 || ^7.0 -> your PHP version (8.0.0-dev) does not satisfy that requirement.
    - laminas/laminas-cache 2.8.1 requires php ^5.6 || ^7.0 -> your PHP version (8.0.0-dev) does not satisfy that requirement.
    - laminas/laminas-cache 2.8.0 requires php ^5.6 || ^7.0 -> your PHP version (8.0.0-dev) does not satisfy that requirement.
    - laminas/laminas-cache 2.7.2 requires php ^5.5 || ^7.0 -> your PHP version (8.0.0-dev) does not satisfy that requirement.
    - laminas/laminas-cache 2.7.1 requires php ^5.5 || ^7.0 -> your PHP version (8.0.0-dev) does not satisfy that requirement.
    - laminas/laminas-cache 2.7.0 requires php ^5.5 || ^7.0 -> your PHP version (8.0.0-dev) does not satisfy that requirement.
    - laminas/laminas-cache 2.6.1 requires php ^5.5 || ^7.0 -> your PHP version (8.0.0-dev) does not satisfy that requirement.
    - Installation request for laminas/laminas-cache ^2.6.1 -> satisfiable by laminas/laminas-cache[2.6.1, 2.7.0, 2.7.1, 2.7.2, 2.8.0, 2.8.1, 2.8.2, 2.8.3, 2.9.0].
```

boesing:
@jfernancordova As you can see, the tests are still failing. This is due to the fact that there are some strange circulare dependencies going on in this package.

Please run `composer test` locally. You will see several problems which needs to get fixed (as there are missing methods such as `assertAttributeSame` and deprecations for prophecy).

You can use `phpspec/prophecy-phpunit` as a dev dependency and add the trait to the tests which are using the `prophesize` method.
To resolve dependencies, you can temporarily add `"version": "2.14.0",` to the `composer.json` (but please do not commit that as this is only needed locally). When you've added `phpspec/prophecy-phpunit` you can remove the `version` property from `composer.json` as the dependency tree should be resolved locally.

@Ocramius @weierophinney Any ideas how to solve such circular dependencies in the future? Is there a way of adding the `version` property to `composer.json` from within the `automatic-releases`?

jfernancordova:
> @jfernancordova As you can see, the tests are still failing. This is due to the fact that there are some strange circulare dependencies going on in this package.
> 
> Please run `composer test` locally. You will see several problems which needs to get fixed (as there are missing methods such as `assertAttributeSame` and deprecations for prophecy).
> 
> You can use `phpspec/prophecy-phpunit` as a dev dependency and add the trait to the tests which are using the `prophesize` method.
> To resolve dependencies, you can temporarily add `"version": "2.14.0",` to the `composer.json` (but please do not commit that as this is only needed locally). When you've added `phpspec/prophecy-phpunit` you can remove the `version` property from `composer.json` as the dependency tree should be resolved locally.
> 
> @Ocramius @weierophinney Any ideas how to solve such circular dependencies in the future? Is there a way of adding the `version` property to `composer.json` from within the `automatic-releases`?

Ok. let me check.

jfernancordova:
@boesing, I fixed all tests and executed `composer test` locally, however, I'm seeing the Build from travis CI, and other dependencies such as: laminas/laminas-http are not working with laminas/laminas-validator, can you suggest any solution for this?, thx

boesing:
@jfernancordova You can add `"version": "2.14.0"` to `composer.json` in a single commit. 
After that, push to your branch. Let travis do its job. 
- When everything is green, *hooray*. Copy the link to the test-results and paste them in here.
- If not, try to determine the problems and fix them and post the link to the green test-results here.

When done, drop the commit from history and force-push back to your branch 👍 

Not sure if there is a better way but thats what comes to my mind.

jfernancordova:
> @jfernancordova You can add `"version": "2.14.0"` to `composer.json` in a single commit.
> After that, push to your branch. Let travis do its job.
> 
>     * When everything is green, _hooray_. Copy the link to the test-results and paste them in here.
> 
>     * If not, try to determine the problems and fix them and post the link to the green test-results here.
> 
> 
> When done, drop the commit from history and force-push back to your branch +1
> 
> Not sure if there is a better way but thats what comes to my mind.

Hi @boesing, this is the link: https://travis-ci.com/github/laminas/laminas-validator/builds/190545033, I hope it's ok.

jfernancordova:
@boesing what do you think?, I just changed some code and fixed all tests, anything else?

test/Db/NoRecordExistsTest.php
```diff
@@ -204,6 +204,14 @@ public function testThrowsExceptionWithNoAdapter()
      */
     public function testWithSchema()
     {
+        if (explode('.', PHP_VERSION)[0] >= 8.0) {
```
froschdesign: [PHP's `version_compare`](https://www.php.net/manual/function.version-compare.php) can be used here.

test/Db/NoRecordExistsTest.php
```diff
@@ -218,6 +226,14 @@ public function testWithSchema()
      */
     public function testWithSchemaNoResult()
     {
+        if (explode('.', PHP_VERSION)[0] >= 8.0) {
```
froschdesign: The same here and so on…

test/InArrayTest.php
```diff
@@ -142,10 +142,10 @@ public function testStrictComparisons()
         $this->assertTrue($validator->getStrict());
         $this->assertFalse($validator->isValid('b'));
         $this->assertFalse($validator->isValid('a'));
-        $this->assertTrue($validator->isValid('A'));
+        $this->assertFalse($validator->isValid('A'));
```
froschdesign: This means a change of logic. This cannot be correct and should be checked again.

test/InArrayTest.php
```diff
@@ -142,10 +142,10 @@ public function testStrictComparisons()
...
         $this->assertFalse($validator->isValid('0'));
         $this->assertFalse($validator->isValid('1a'));
-        $this->assertTrue($validator->isValid(0));
+        $this->assertFalse($validator->isValid(0));
```
froschdesign: The same here. The haystack contains the following values:

https://github.com/laminas/laminas-validator/blob/83c2a967cd823a76a81589e895363aac73b4d646/test/InArrayTest.php#L133-L137

test/Db/NoRecordExistsTest.php
jfernancordova: Great, I'll fix these conditionals, thx.

test/InArrayTest.php
jfernancordova: You're right, I will check this.

test/ValidatorPluginManagerFactoryTest.php
```diff
@@ -28,10 +31,7 @@ public function testFactoryReturnsPluginManager()
 
         if (method_exists($validators, 'configure')) {
             // laminas-servicemanager v3
-            $this->assertAttributeSame($container, 'creationContext', $validators);
-        } else {
-            // laminas-servicemanager v2
-            $this->assertSame($container, $validators->getServiceLocator());
+            $this->assertObjectHasAttribute('creationContext', $validators);
```
froschdesign: This section is no longer useful and testing internal implementation details is the wrong way.

src/InArray.php
```diff
@@ -193,8 +193,16 @@ public function isValid($value)
...
+    
+                    if ($el == $value && self::COMPARE_STRICT != $this->strict) {
+                        if (InArray::COMPARE_NOT_STRICT_AND_PREVENT_STR_TO_INT_VULNERABILITY == $this->strict) {
+                            if ($this->inArrayRecursive($el, $haystack)) {
+                                return true;
+                            }
+                            return false;
+                        }
+                        return true;
+                    }elseif (InArray::COMPARE_NOT_STRICT == $this->strict){
```
froschdesign: If you find a problem in this the validator independent of the PHP 8 support then create a separate pull request with an unit test. Otherwise, please revert the changes in this file.  Thanks!

test/Db/NoRecordExistsTest.php
glensc: if you are comparing the php version, use PHP_VERSION_ID constant

glensc: altho, you should use the requires annotation here instead:


```
     * @requires PHP < 8.0
```

- https://phpunit.readthedocs.io/en/9.3/incomplete-and-skipped-tests.html#skipping-tests-using-requires

jfernancordova: Great, thank you @glensc!

test/ValidatorPluginManagerFactoryTest.php
jfernancordova: Ok, I'll remove the condition

src/InArray.php
jfernancordova: Ok, I revert this change and add a version compare in the test, because only 2 or 3 cases are not working with PHP 8.0

froschdesign: @jfernancordova
> …add a version compare in the test, because only 2 or 3 cases are not working with PHP 8.0.

I am afraid there is a misunderstanding here. Sorry!
All cases must be passed otherwise it would be a break of the behaviour. The validator must produce the same results in all PHP versions.

jfernancordova: I see, @froschdesign so if I revert these changes, the tests aren't going to work correctly. I'll try to fix this in this PR first.

froschdesign: Thanks in advance! 👍 👍 👍

jfernancordova:
@froschdesign I analyzed the code again. I could understand the complexity, so I did a refactor in which the code is cleaner now. What do you think?. Any suggestions are welcome.

src/AbstractValidator.php
```diff
@@ -132,11 +132,11 @@ public function setOptions($options = [])
         foreach ($options as $name => $option) {
             $fname = 'set' . ucfirst($name);
             $fname2 = 'is' . ucfirst($name);
-            if (($name !== 'setOptions') && method_exists($this, $name)) {
+            if (($name !== 'setOptions') && get_parent_class($this) && method_exists($this, $name)) {
```
boesing: Could you explain why you've added these additional checks with `get_parent_class`?
As this is an abstract class, this check is *always* true or do I miss something here?

In case we do not need this additional check (and I think so), please revert this change.

src/InArray.php
```diff
@@ -216,14 +220,32 @@ public function isValid($value)
...
+            
         }
 
         $this->error(self::NOT_IN_ARRAY);
         return false;
     }
+    
+    private function isMultiArray($array)
+    {
+        $rv = array_filter($array, 'is_array');
```
boesing: depending on the size of the given array, you might want to use `foreach` with early return instead. This will avoid unnecessary CPU cycles.

src/InArray.php
```diff
@@ -216,14 +220,32 @@ public function isValid($value)
...
                 return true;
             }
+            
         }
 
         $this->error(self::NOT_IN_ARRAY);
         return false;
     }
+    
+    private function isMultiArray($array)
```
boesing: Please also add proper typehints to the methods parameter and return value.

boesing (changes requested):

src/AbstractValidator.php
jfernancordova: Yes, I just considered an extra validation, but you're right, I'm going to revert this change.

.travis.yml
```diff
@@ -13,21 +13,6 @@ env:
 matrix:
   fast_finish: true
   include:
-    - php: 7.1
-      env:
-        - DEPS=lowest
-    - php: 7.1
-      env:
-        - DEPS=latest
-        - EXECUTE_HOSTNAME_CHECK=true
```
boesing, line 22: Oh, as far as I can see here, we dropped some checks.
Could you re-add these to PHP 7.4?
Sorry, realized just yet.
After this, we should be good to go (I think the CS check will find some problems, maybe you check yourself by locally executing `composer cs-check`).

boesing (changes requested):

.travis.yml
jfernancordova: No problem!

.travis.yml
```diff
@@ -40,6 +25,15 @@ matrix:
     - php: 7.4
       env:
         - DEPS=latest
+        - EXECUTE_HOSTNAME_CHECK=true
```
boesing, line 28: `TEST_COVERAGE` and `CS_CHECK` are missing in here aswell.

boesing (changes requested):

.travis.yml
jfernancordova: Uh, I'll add them too, sorry.

jfernancordova:
@boesing Adding `TEST_COVERAGE` is returning an error like: `PHP Fatal error:  Uncaught Error: Maximum function nesting level of '256' reached, aborting! in /home/travis/build/laminas/laminas-validator/vendor/nikic/php-parser/lib/PhpParser/NodeAbstract.php`. I think it's about of `php.ini` configuration file.

.travis.yml
jfernancordova: @boesing Adding TEST_COVERAGE is returning an error: PHP Fatal error: Uncaught Error: Maximum function nesting level of '256' reached, aborting! in /home/travis/build/laminas/laminas-validator/vendor/nikic/php-parser/lib/PhpParser/NodeAbstract.php. I think it's about of php.ini or xdebug.ini configuration file. I modified the travis yml, on the before_install section `echo "xdebug.max_nesting_level=3000" >> xdebug.ini`

.travis.yml
```diff
@@ -40,8 +25,21 @@ matrix:
...
+        - DEPS=lowest
+        - COMPOSER_ARGS="--ignore-platform-reqs"
+    - php: nightly
+      env:
+        - DEPS=latest
+        - COMPOSER_ARGS="--ignore-platform-reqs"
 
 before_install:
+  - export INI=~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
+  - echo xdebug.max_nesting_level = 3000 >> $INI
```
boesing: Does the error for which this change is necessary happen for coverage only? 
I'd prefer having this within `phpunit.xml.dist` instead as its phpunit related and could be a problem in local development aswell.

See phpunit configuration documentation: https://phpunit.readthedocs.io/en/9.3/configuration.html#the-ini-element

composer.json
```diff
@@ -2,6 +2,7 @@
     "name": "laminas/laminas-validator",
     "description": "Validation classes for a wide range of domains, and the ability to chain validators to create complex validation criteria",
     "license": "BSD-3-Clause",
+    "version": "2.14.0",
```
boesing: after tests are green, please remove this from the `composer.json` by removing the related commit.

boesing (changes requested):

.travis.yml
jfernancordova: Yes, only for coverage, ok, I'll check about it.

composer.json
jfernancordova: Ok

jfernancordova:
@boesing Before removing the composer's version, all tests are green!
https://travis-ci.com/github/laminas/laminas-validator/builds/197589684

boesing (approved):
LGTM

samsonasik:
please update laminas-http require-dev to ^2.14

jfernancordova:
> please update laminas-http require-dev to ^2.14

Ok!

src/InArray.php
```diff
@@ -213,17 +217,35 @@ public function isValid($value)
             ) {
                 foreach ($haystack as &$h) {
                     if (is_int($h) || is_float($h)) {
-                        $h = (string) $h;
+                        $h = (string)$h;
                     }
                 }
+
+                if (! $this->isMultiArray($haystack) && array_key_exists($value, array_flip($haystack))) {
```
Ocramius: Is flipping safe here? What of the array contains objects?

src/InArray.php
```diff
@@ -213,17 +217,35 @@ public function isValid($value)
...
+                    return true;
+                }
+            }
+
+            if (array_search($value, $haystack, self::COMPARE_STRICT == $this->strict)) {
```
Ocramius: Can `==` and `!=` be replaced with strict comparison (and is it safe to do so)?

src/InArray.php
```diff
@@ -213,17 +217,35 @@ public function isValid($value)
...
+            if (self::COMPARE_NOT_STRICT == $this->strict) {
                 return true;
             }
         }
 
         $this->error(self::NOT_IN_ARRAY);
         return false;
     }
+
+    private function isMultiArray(array $array): bool
```
Ocramius: s/multi/multidimensional?

test/Db/NoRecordExistsTest.php
```diff
@@ -200,6 +200,10 @@ public function testThrowsExceptionWithNoAdapter()
     /**
      * Test that schemas are supported and run without error
      *
+     * @requires PHP < 8.0
+     * For PHP 8.0 Library laminas/laminas-stdlib/src/ArrayUtils.php needs to be upgraded
```
Ocramius: This means that we aren't yet php8-compatible, and we should fix that dependency (bug) first

test/Db/NoRecordExistsTest.php
```diff
@@ -214,6 +218,10 @@ public function testWithSchema()
     /**
      * Test that schemas are supported and run without error
      *
+     * @requires PHP < 8.0
```
Ocramius: This means that we aren't yet php8-compatible, and we should fix that dependency (bug) first

Ocramius (changes requested):
We really can't use `@requires php<8` for releasing this: if it doesn't run, it isn't ready 😞

.travis.yml
```diff
@@ -40,6 +25,17 @@ matrix:
     - php: 7.4
       env:
         - DEPS=latest
+        - EXECUTE_HOSTNAME_CHECK=true
+        - CS_CHECK=true
+        - TEST_COVERAGE=true
+    - php: nightly
```
froschdesign: Version 8.0 should be used here.

src/InArray.php
jfernancordova: Hi @Ocramius, it was the solution I thought could work and all tests passed: https://travis-ci.com/github/laminas/laminas-validator/builds/197589684, unfortunately, the last implementation breaks with PHP 8, if you have any suggestions, let me know, please!

However, I could reevaluate it!

src/InArray.php
jfernancordova: Sorry @Ocramius, I don't understand you, do you want to replace it with strict comparison or know if it safe?

src/InArray.php
jfernancordova: yes multidimensional!

test/Db/NoRecordExistsTest.php
jfernancordova: That's right!

test/Db/NoRecordExistsTest.php
jfernancordova: That's right!

src/InArray.php
glensc: @richardjennings @Ocramius suggested here you to change the word `multi` with `multidimensional`

src/InArray.php
glensc: @jfernancordova, I read it as to make a change to do strict compare but validate first is the change legal first, i.e not introducing a bug by making such change.

src/InArray.php
jfernancordova: I get it now, thanks @glensc!

.travis.yml
```diff
@@ -13,33 +13,33 @@ env:
...
       env:
         - DEPS=latest
-    - php: 7.4
+        - COMPOSER_ARGS="--ignore-platform-reqs"
+    - php: 8.0
       env:
         - DEPS=lowest
-    - php: 7.4
+        - COMPOSER_ARGS="--ignore-platform-reqs"
+    - php: nightly
```
Ocramius: If we test nightly, we should probably allow failures there (it's very unstable)

test/ValidatorPluginManagerFactoryTest.php
```diff
@@ -15,24 +15,18 @@
...
+    use ProphecyTrait;
     public function testFactoryReturnsPluginManager()
     {
         $container = $this->prophesize(ContainerInterface::class)->reveal();
         $factory = new ValidatorPluginManagerFactory();
 
         $validators = $factory($container, ValidatorPluginManagerFactory::class);
         $this->assertInstanceOf(ValidatorPluginManager::class, $validators);
-
-        if (method_exists($validators, 'configure')) {
```
Ocramius, line 29: Ìsn't v3 "current"?

src/InArray.php
jfernancordova: I'm working on it!

Ocramius (changes requested):

src/InArray.php
jfernancordova: @Ocramius what do you think about this?
```
$haystack = is_object($haystack) ? get_object_vars($haystack) : $haystack;
if (! $this->isMultidimensionalArray($haystack) && array_key_exists($value, array_flip($haystack))) {
    return true;
}
```

.travis.yml
jfernancordova: so I remove nightly, right?

samsonasik: iirc, nightly was for php 8 dev, I think it can be removed.

test/ValidatorPluginManagerFactoryTest.php
jfernancordova: @Ocramius I tried to set these asserts, especially v3, and they're deprecated, that's why I remove this block.

.travis.yml
jfernancordova: OK!

lahivee:
Hey guys, is there any way to push this through donations? We are depending on this package to support PHP8 for this package: https://github.com/beyondcode/expose/pull/163

danmichaelo:
Per https://github.com/laminas/laminas-validator/pull/79#discussion_r550776411 , can we remove `laminas/laminas-cache` as a dev dependency? Couldn't find a single reference to caches in this repo: `find . -iname '*.php' -exec grep -i 'Cache' {} \;`

composer.json
```diff
@@ -36,13 +36,14 @@
         "laminas/laminas-config": "^2.6",
         "laminas/laminas-db": "^2.7",
         "laminas/laminas-filter": "^2.6",
-        "laminas/laminas-http": "^2.5.4",
+        "laminas/laminas-http": "^2.14",
```
weierophinney: Bump to 2.14.2, please

src/InArray.php
```diff
@@ -173,7 +173,7 @@ public function isValid($value)
         // we type cast the input to a string
         if (self::COMPARE_NOT_STRICT_AND_PREVENT_STR_TO_INT_VULNERABILITY == $this->strict
             && (is_int($value) || is_float($value))) {
-            $value = (string) $value;
+            $value = (string)$value;
         }
```
weierophinney: These changes are all unrelated... and I suspect you ran a CS tool over this other than our own `composer cs-check` and `composer cs-fix` (which use phpcs and the laminas-coding-standard). Please either revert these changes, or run `composer cs-fix` and commit the updates.

src/InArray.php
```diff
@@ -191,10 +191,14 @@ public function isValid($value)
...
-                        $el = (string) $el;
+                        $el = (string)$el;
                     }
 
-                    if ($el == $value) {
+                    if ($el == $value && self::COMPARE_STRICT != $this->strict) {
+                        return true;
+                    }
+
+                    if (self::COMPARE_NOT_STRICT == $this->strict) {
```
weierophinney: What is this change for, exactly? It seems unrelated to the bump to PHP 8...

src/InArray.php
```diff
@@ -213,17 +217,45 @@ public function isValid($value)
...
                     }
                 }
+
+                if (! $this->isMultidimensionalArray($haystack) && $this->isArrayKeyExists($value, $haystack)) {
+                    return true;
+                }
             }
 
-            if (in_array($value, $haystack, self::COMPARE_STRICT == $this->strict)) {
+            if (array_search($value, $haystack, self::COMPARE_STRICT == $this->strict)) {
```
weierophinney: Please revert this change; they are not semantically the same!

We want to see if the value is in the array, not return the first match. `in_array()` is the correct one to use here.

src/InArray.php
```diff
@@ -213,17 +217,45 @@ public function isValid($value)
...
+    {
+        foreach ($array as $value) {
+            if (is_array($value)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private function isArrayKeyExists(string $key, array $haystack): bool
```
weierophinney: This is misnamed, and duplicates the functionality of `in_array()`. Please remove, and replace calls to this method with `in_array($value, $haystack)`.

src/InArray.php
```diff
@@ -213,17 +217,45 @@ public function isValid($value)
             ) {
                 foreach ($haystack as &$h) {
                     if (is_int($h) || is_float($h)) {
-                        $h = (string) $h;
+                        $h = (string)$h;
                     }
                 }
+
+                if (! $this->isMultidimensionalArray($haystack) && $this->isArrayKeyExists($value, $haystack)) {
```
weierophinney: This is unrelated to updating to PHP 8. Please remove.

weierophinney (changes requested):
There are a lot of CS changes that are (a) unrelated to updating to PHP 8, and (b) incorrect by our coding standards. Additionally, the InArray validator has a number of changes outside the scope of the PR (adding PHP 8 support), and without new tests to validate the new functionality.

Please run `composer cs-fix`, then `composer cs-check`, fixing and committing any changes needed to make the checks pass. Also, please back out the changes in the `InArray` validator; if you want to see them in a future version, please submit them as a separate PR, with tests demonstrating and validating the changes.

jfernancordova:
Hi @weierophinney, As you can see, reverting the changes from inArray.php class, break the tests with PHP 8.0, https://travis-ci.com/github/laminas/laminas-validator/jobs/468944524, that's why I changed the code, the native function in_array seems not working as the previous versions.

However, I'm going to create another PR with the new changes and their tests for inArray validator.

src/InArray.php
jfernancordova: Ok!

lahivee:
@jfernancordova thank you so much for your work on this!

weierophinney:
@jfernancordova —

> reverting the changes from inArray.php class, break the tests with PHP 8.0, https://travis-ci.com/github/laminas/laminas-validator/jobs/468944524, that's why I changed the code, the native function in_array seems not working as the previous versions.
> 
> However, I'm going to create another PR with the new changes and their tests for inArray validator.

I'm fine with including the changes here; in fact, since the PHP 8 tests _fail_ without them, then we need them before merging. I just didn't understand why you weren't using `in_array()` directly, and there were no comments in the code to indicate why.

Speaking of: it may be good to create a bug on bugs.php.net about this. The PHP 8 changelog does not mention anything about the change in behavior, and I don't see any bugs currently filed around it.

jfernancordova:
> @jfernancordova —
> 
> > reverting the changes from inArray.php class, break the tests with PHP 8.0, https://travis-ci.com/github/laminas/laminas-validator/jobs/468944524, that's why I changed the code, the native function in_array seems not working as the previous versions.
> > However, I'm going to create another PR with the new changes and their tests for inArray validator.
> 
> I'm fine with including the changes here; in fact, since the PHP 8 tests _fail_ without them, then we need them before merging. I just didn't understand why you weren't using `in_array()` directly, and there were no comments in the code to indicate why.
> 
> Speaking of: it may be good to create a bug on bugs.php.net about this. The PHP 8 changelog does not mention anything about the change in behavior, and I don't see any bugs currently filed around it.

You're right, I did not explain why I removed the code, especially why I did not use the `in_array()` function.
Would you like me to fix InArray validator class adding why I don't use the `in_array()` in this PR?
It would be the same before reverting the requested changes.

weierophinney:
> Would you like me to fix InArray validator class adding why I don't use the `in_array()` in this PR?
> It would be the same before reverting the requested changes.

Yes, please!

jfernancordova:
Hi again @weierophinney! I started to modify the inArray validator, I realized that I did not need some functions I added, just changed the way of conditionals and using `in_array()` function, I ask you again if you want to explain this way but I think it is implicit.

src/InArray.php
```diff
@@ -216,9 +220,17 @@ public function isValid($value)
...
+            }
+
+            if (array_search($value, $haystack, self::COMPARE_STRICT == $this->strict)) {
+                return true;
             }
 
-            if (in_array($value, $haystack, self::COMPARE_STRICT == $this->strict)) {
+            if (self::COMPARE_NOT_STRICT == $this->strict) {
                 return true;
             }
```
weierophinney: This part (and the related comparison in the recursive comparison section)... seems wrong. You're testing to see if a constant value referencing "not strict comparisons" matches the  flag provided by the consumer as to whether or not to do strict comparisons. You're not comparing the value **at all**, and yet if these match, you're returning true.

At best, this is only **accidentally** causing tests to pass. At worse, it's hiding something else.

So, I did some digging, and had quite a surprise: The behavior of `==` appears to have changed in PHP 8.

I did the following:

```php
$values = ['test', 0, 'A', 0.0, 'foo', 1, 'a', 'c'];
foreach ($values as $value) {
    printf("- Is b equal to %s?\t%s\n", $value, 'b' == $value ? 'Yes' : 'No');
}
```

Under PHP 7 versions, you get the following output:

```text
- Is b equal to test?	No
- Is b equal to 0?	Yes
- Is b equal to A?	No
- Is b equal to 0?	Yes
- Is b equal to foo?	No
- Is b equal to 1?	No
- Is b equal to a?	No
- Is b equal to c?	No
```

However, under PHP 8, we get:

```text
- Is b equal to test?	No
- Is b equal to 0?	No
- Is b equal to A?	No
- Is b equal to 0?	No
- Is b equal to foo?	No
- Is b equal to 1?	No
- Is b equal to a?	No
- Is b equal to c?	No
```

This is true even with strict_types off. And, obviously, this is hugely problematic for this particular comparison.

Looking in the [PHP 8 upgrade guide](https://github.com/php/php-src/blob/PHP-8.0/UPGRADING), there is a note about this:

> Non-strict comparisons between numbers and non-numeric strings now work by casting the number to string and comparing the strings. Comparisons between numbers and numeric strings continue to work as before. Notably, this means that `0 == "not-a-number"` is considered false now.

As such, I'm going to update the tests to vary the assertions based on PHP version, and I'll push the changes back to your branch.

jfernancordova: Oh I see, I'm sorry for not realizing about this, I just trusted in the tests, thank you for the details, very appreciate @weierophinney

jfernancordova:
It was an amazing experience! Thanks for everything @boesing @glensc @weierophinney @Ocramius

lahivee:
thank you for your work on PHP8 support! when do you plan to release this?

weierophinney:
There's a few other PRs marked for the 2.14.0 release that I'm reviewing and helping complete now; hopefully later today, but it may be tomorrow.

PR #77, opened 2020-12-17
add heic to valid isImage mime checks

christaggart:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

  - Due to mainstream popularization of HEIC as the default image format for iOS devices, adding image/heic as a default supported mimetype is useful here
  - What will it enable?
  - support for heic images when tested by the isImage validator

### DIFF
--- a/src/File/IsImage.php
+++ b/src/File/IsImage.php
@@ -59,6 +59,8 @@ class IsImage extends MimeType
             'image/cgm',
             'image/fif',
             'image/gif',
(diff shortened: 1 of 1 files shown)


christaggart:
signed the commit but DCO doesn't look like it re-ran

PR #89, opened 2021-01-23
Fix Type Error when validating a `DateTimeImmutable` instance to the `DateStep` validator

zluiten:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

The `DateStep` validator fails when providing a `DateTimeImmutable` instance since the `fallbackIncrementalIterationLogic` function typehints to `DateTime`. This PR primarily fixes that issue.

Besides that I've added several typing improvements regarding `DateTime` instances and other primitives on the `Date` validator (which `DateStep` extends).
Meanwhile fixed a bunch of psalm errors as well.

PR #91, opened 2021-01-24
Fix immutability issue when validating a `DateTimeImmutable` instance with the `DateStep` validator

zluiten:
See #89, but this one actually fixes the bug.

### DIFF
--- a/src/DateStep.php
+++ b/src/DateStep.php
@@ -366,13 +366,13 @@ class DateStep extends Date
         $dateModificationOperation           = $isIncrementalStepping ? 'add' : 'sub';
 
         for ($offsetIterations = 0; $offsetIterations < $requiredIterations; $offsetIterations += 1) {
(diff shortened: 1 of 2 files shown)


src/DateStep.php
```diff
@@ -366,13 +366,13 @@ private function fallbackIncrementalIterationLogic(
         $dateModificationOperation           = $isIncrementalStepping ? 'add' : 'sub';
 
         for ($offsetIterations = 0; $offsetIterations < $requiredIterations; $offsetIterations += 1) {
-            $baseDate->{$dateModificationOperation}($minimumInterval);
+            $baseDate = $baseDate->{$dateModificationOperation}($minimumInterval);
```
Ocramius, line 369: Well, yeah, #89 fixed the type issue, but obviously not the validation issue :D

Ocramius (approved):
LGTM, thanks for the quick follow-up! 👍

PR #93, opened 2021-01-25
Fix psalm errors on Date and DateStep validator

zluiten:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

Since I worked on the Date and DateStep validator I did an attempt to fix (most) psalm errors for those validators:

70 down, many to go.

I'll explain some of the changes by adding comments.

```
Scanning files...
Analyzing files...

░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░  60 / 197 (30%)
░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 120 / 197 (60%)
░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 180 / 197 (91%)
░░░░░░░░░░░░░░░░░
------------------------------
70 errors fixed

------------------------------
No errors found!
------------------------------
1940 other issues found.
You can display them with --show-info=true
------------------------------
Psalm can automatically fix 7 of these issues.
Run Psalm again with 
--alter --issues=InvalidReturnType --dry-run
to see what it can fix.
------------------------------

Checks took 12.62 seconds and used 287.018MB of memory
Psalm was able to infer types for 94.7835% of the codebase

```

### DIFF
--- a/src/Date.php
+++ b/src/Date.php
@@ -10,7 +10,14 @@ namespace Laminas\Validator;
 
 use DateTime;
 use DateTimeImmutable;
+use DateTimeInterface;
 use Traversable;
+use function array_shift;
+use function func_get_args;
+use function gettype;
+use function implode;
+use function is_array;
+use function iterator_to_array;
 
 /**
  * Validates that a given value is a DateTime instance or can be converted into one.
@@ -81,7 +88,7 @@ class Date extends AbstractValidator
     /**
      * Returns the format option
      *
-     * @return string|null
+     * @return string
      */
     public function getFormat()
     {
@@ -94,7 +101,7 @@ class Date extends AbstractValidator
      * Format cannot be null.  It will always default to 'Y-m-d', even
      * if null is provided.
      *
-     * @param  string $format
+     * @param string|null $format
      * @return $this provides a fluent interface
      * @todo   validate the format
      */
@@ -104,21 +111,28 @@ class Date extends AbstractValidator
         return $this;
     }
 
+    /**
+     * @param bool $strict
+     * @return $this
+     */
     public function setStrict(bool $strict) : self
     {
         $this->strict = $strict;
         return $this;
     }
 
+    /**
+     * @return bool
+     */
     public function isStrict() : bool
     {
         return $this->strict;
     }
 
     /**
-     * Returns true if $value is a DateTime instance or can be converted into one.
+     * Returns true if $value is a DateTimeInterface instance or can be converted into one.
      *
-     * @param  string|array|int|DateTime|DateTimeImmutable $value
+     * @param  string|numeric|array|DateTimeInterface $value
      * @return bool
      */
     public function isValid($value)
@@ -142,48 +156,59 @@ class Date extends AbstractValidator
     /**
      * Attempts to convert an int, string, or array to a DateTime object
      *
-     * @param  string|int|array $param
-     * @param  bool             $addErrors
-     * @return bool|DateTime
+     * @param  string|numeric|array|DateTimeInterface   $param
+     * @param  bool                                     $addErrors
+     * @return false|DateTime
      */
     protected function convertToDateTime($param, $addErrors = true)
     {
-        if ($param instanceof DateTime || $param instanceof DateTimeImmutable) {
+        if ($param instanceof DateTime) {
             return $param;
         }
 
-        $type = gettype($param);
-        if (! in_array($type, ['string', 'integer', 'double', 'array'])) {
-            if ($addErrors) {
-                $this->error(self::INVALID);
-            }
-            return false;
+        if ($param instanceof DateTimeImmutable) {
+            return DateTime::createFromImmutable($param);
         }
 
-        $convertMethod = 'convert' . ucfirst($type);
-        return $this->{$convertMethod}($param, $addErrors);
+        $type = gettype($param);
+        switch ($type) {
+            case 'string':
+                return $this->convertString($param, $addErrors);
+            case 'integer':
+                return $this->convertInteger($param);
+            case 'double':
+                return $this->convertDouble($param);
+            case 'array':
+                return $this->convertArray($param, $addErrors);
+        }
+
+        if ($addErrors) {
+            $this->error(self::INVALID);
+        }
+
+        return false;
     }
 
     /**
      * Attempts to convert an integer into a DateTime object
      *
      * @param  integer $value
-     * @return bool|DateTime
+     * @return false|DateTime
      */
     protected function convertInteger($value)
     {
-        return date_create("@$value");
+        return DateTime::createFromFormat('U', (string) $value);
     }
 
     /**
      * Attempts to convert an double into a DateTime object
      *
      * @param  double $value
-     * @return bool|DateTime
+     * @return false|DateTime
      */
     protected function convertDouble($value)
     {
-        return DateTime::createFromFormat('U', $value);
+        return DateTime::createFromFormat('U', (string) $value);
     }
 
     /**
@@ -191,7 +216,7 @@ class Date extends AbstractValidator
      *
      * @param  string $value
      * @param  bool   $addErrors
-     * @return bool|DateTime
+     * @return false|DateTime
      */
     protected function convertString($value, $addErrors = true)
     {
@@ -215,7 +240,7 @@ class Date extends AbstractValidator
      *
      * @param  array $value
      * @param  bool  $addErrors
-     * @return bool|DateTime
+     * @return false|DateTime
      * @todo   enhance the implosion
      */
     protected function convertArray(array $value, $addErrors = true)
--- a/src/Date.php
+++ b/src/Date.php
@@ -6,12 +6,15 @@
  * @license   https://github.com/laminas/laminas-validator/blob/master/LICENSE.md New BSD License
  */
 
+declare(strict_types=1);
+
 namespace Laminas\Validator;
 
 use DateTime;
 use DateTimeImmutable;
 use DateTimeInterface;
 use Traversable;
+
 use function array_shift;
 use function func_get_args;
 use function gettype;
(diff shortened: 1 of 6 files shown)


src/Date.php
```diff
@@ -94,7 +101,7 @@ public function getFormat()
      * Format cannot be null.  It will always default to 'Y-m-d', even
      * if null is provided.
      *
-     * @param  string $format
+     * @param string|null $format
```
zluiten, line 99: Added this one since tests exist which test this. The setFormat function changes `$format` to the default value in case it is `empty`.

src/Date.php
```diff
@@ -104,21 +111,28 @@ public function setFormat($format = self::FORMAT_DEFAULT)
         return $this;
     }
 
+    /**
+     * @param bool $strict
+     * @return $this
+     */
```
froschdesign: The same information is already contained in the method signature, so the DocBlock can be omitted.

src/Date.php
```diff
@@ -104,21 +111,28 @@ public function setFormat($format = self::FORMAT_DEFAULT)
...
     {
         return $this->strict;
     }
 
     /**
-     * Returns true if $value is a DateTime instance or can be converted into one.
+     * Returns true if $value is a DateTimeInterface instance or can be converted into one.
      *
-     * @param  string|array|int|DateTime|DateTimeImmutable $value
+     * @param  string|numeric|array|DateTimeInterface $value
```
zluiten, line 123: `numeric` because floats are accepted as well. Arguably `int|float` might be better since numeric also means numerical strings which are not really supported. In case of a numerical string an attempt to create a DateTime instance will be performed based on the configured format by the `convertToString` function.

src/Date.php
```diff
@@ -142,56 +156,67 @@ public function isValid($value)
...
      * Attempts to convert an integer into a DateTime object
      *
      * @param  integer $value
-     * @return bool|DateTime
+     * @return false|DateTime
      */
     protected function convertInteger($value)
     {
-        return date_create("@$value");
+        return DateTime::createFromFormat('U', (string) $value);
```
zluiten, line 188: For consistency with `convertDouble`...

src/DateStep.php
```diff
@@ -209,13 +232,14 @@ public function isValid($value)
         $step = $this->getStep();
 
         // Same date?
-        if ($valueDate == $baseDate) {
+        if ($valueDate === $baseDate) {
```
zluiten: Check can be safely explicit because `Date::convertToDateTime` makes sure both $valueDate and $baseDate will be of type `DateTime` even when one of them was given as a DateTimeImmutable.

src/DateStep.php
```diff
@@ -209,13 +232,14 @@ public function isValid($value)
...
             return true;
         }
 
         // Optimization for simple intervals.
         // Handle intervals of just one date or time unit.
         $intervalParts = explode('|', $step->format('%y|%m|%d|%h|%i|%s'));
+        $intervalParts = array_map('intval', $intervalParts);
```
zluiten, line 249: Initial contents of $intervalParts are numerical strings, but typehints of called functions point to and count on integers.

src/DateStep.php
```diff
@@ -225,17 +249,17 @@ public function isValid($value)
...
+        $timeDiff  = $absoluteValueDate->diff($absoluteBaseDate, true);
+        $diffParts = array_combine($unitKeys, array_map('intval', explode('|', $timeDiff->format('%y|%m|%d|%h|%i|%s'))));
 
-        if (5 === $partCounts['0']) {
+        if (5 === $partCounts[0]) {
             // Find the unit with the non-zero interval
-            $intervalUnit = null;
-            $stepValue    = null;
+            $intervalUnit = 'days';
+            $stepValue    = 1;
```
zluiten, line 266: Giving an initial value of the same type of what happens in the next foreach statement prevents Psalm complaining with a `PossiblyNullArgument`.

src/DateStep.php
```diff
@@ -225,17 +249,17 @@ public function isValid($value)
...
             foreach ($intervalParts as $key => $value) {
-                if (0 != $value) {
+                if (0 !== $value) {
```
zluiten, line 268: Now we can do explicit checks since `$value` is always of type int.

src/DateStep.php
```diff
@@ -345,34 +369,49 @@ public function isValid($value)
...
+     * @throws Exception\InvalidArgumentException
      */
     private function fallbackIncrementalIterationLogic(
         DateTimeInterface $baseDate,
         DateTimeInterface $valueDate,
         array $intervalParts,
         array $diffParts,
         DateInterval $step
-    ) {
+    ) : bool {
```
zluiten: On private functions a return type can be safely added.

src/DateStep.php
```diff
@@ -345,34 +369,49 @@ public function isValid($value)
...
         list($minSteps, $requiredIterations) = $this->computeMinStepAndRequiredIterations($intervalParts, $diffParts);
         $minimumInterval                     = $this->computeMinimumInterval($intervalParts, $minSteps);
         $isIncrementalStepping               = $baseDate < $valueDate;
-        $dateModificationOperation           = $isIncrementalStepping ? 'add' : 'sub';
+
+        if (! ($baseDate instanceof DateTime || $baseDate instanceof DateTimeImmutable)) {
```
zluiten, line 396: This function relies on the `add` and `sub` method of the `DateTime` and `DateTimeImmutable` which are not part of the `DateTimeInterface`.

src/DateStep.php
```diff
@@ -427,7 +466,7 @@ private function computeMinStepAndRequiredIterations(array $intervalParts, array
             $minSteps               = floor($minSteps / $requiredStepIterations);
         }
 
-        return [$minSteps, $minSteps ? $requiredStepIterations : 0];
+        return [(int) $minSteps, $minSteps ? (int) $requiredStepIterations : 0];
```
zluiten, line 471: Both can be cast to integer since the values will not be higher than `PHP_INT_MAX` (see if-statement right above this line).

src/Date.php
zluiten: Is there a style guide that states these kind of agreements?

froschdesign: See: https://docs.laminas.dev/laminas-coding-standard/
 
> ## 9. Commenting and DocBlocks
> They SHOULD NOT be used for already typehinted arguments, except arrays.

https://docs.laminas.dev/laminas-coding-standard/v2/coding-style-guide/#9-commenting-and-docblocks

zluiten: Thanks @froschdesign Must have overlooked it! I'll have a look and apply changes accordingly.

src/DateStep.php
andrew-demb: Strict comparison fail for dates with equal string representation, but they are not the same object in memory - https://3v4l.org/gKB52

test/DateStepTest.php
```diff
@@ -140,20 +150,20 @@ public function testWithDateTimeImmutableType() : void
         $this->assertTrue($validator->isValid(new DateTimeImmutable('1970-01-03T00:00:02Z')));
     }
 
-    public function testGetMessagesReturnsDefaultValue(): void
+    public function testGetMessagesReturnsDefaultValue() : void
```
froschdesign: @netiul 
Please revert all these changes. See:

> ## 4.5 Method and Function Arguments
> When you have a return type declaration present, there MUST be one space after the colon followed by the type declaration. The colon and declaration MUST be on the same line as the argument list closing parenthesis with no spaces between the two characters.

https://docs.laminas.dev/laminas-coding-standard/v2/coding-style-guide/#4-5-method-and-function-arguments

zluiten: Will do. The docs are a bit confusing though. The example here: https://docs.laminas.dev/laminas-coding-standard/v2/coding-style-guide/#1-2-example shows different and the code base mainly uses a space between function declaration and the colon. Personal preference is without space though :)

src/DateStep.php
zluiten: You are absolutely right. Silly me.

test/DateStepTest.php
froschdesign: This is an error in the documentation and must be reported to https://github.com/laminas/laminas-coding-standard/issues

src/Sitemap/Changefreq.php
```diff
@@ -60,9 +60,6 @@ public function isValid($value)
         }
 
         $this->setValue($value);
-        if (! is_string($value)) {
```
zluiten, line 63: For some reason this one was not in the psalm-baseline file yet. It could be removed because the `is_string` check was done twice in the same control flow.

test/DateTest.php
```diff
@@ -135,36 +140,31 @@ public function testDateTimeImmutable(): void
...
         $this->assertTrue($this->validator->setFormat('m Y')->isValid('01 2010'));
         $this->assertFalse($this->validator->setFormat('d/m/Y')->isValid('2008/10/22'));
         $this->assertTrue($this->validator->setFormat('d/m/Y')->isValid('22/10/08'));
         $this->assertFalse($this->validator->setFormat('d/m/Y')->isValid('22/10'));
-        // Omitting the following assertion, as it varies from 5.3.3 to 5.3.11,
-        // and there is no indication in the PHP changelog as to when or why it
-        // may have changed. Leaving for posterity, to indicate original expectation.
-        // $this->assertFalse($this->validator->setFormat('s')->isValid(0));
+        $this->assertTrue($this->validator->setFormat('s')->isValid('00'));
+        $this->assertFalse($this->validator->setFormat('s')->isValid('0'));
```
zluiten, line 167: I've reinstated a similar check as the one that was commented out. I couldn't verify the exact behaviour even with the mentioned PHP versions (5.3.3-5.3.11). My best guess was that the intention was to validate a date '0' formatted with 's' which requires a leading '0', so: '00 '.

PR #103, opened 2021-06-17
Update TLD list according to IANA 2021061501

PowerKiKi:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

This add 76 new TLDs directly copied from http://www.iana.org/domains/root/db


froschdesign:
@PowerKiKi 
Did you use the [script](https://github.com/laminas/laminas-validator/blob/56e1353e358bd003f52511433792b30682136630/bin/update_hostname_validator.php) in the `bin` directory?

PowerKiKi:
Now I did :innocent: 

The script removed a few TLD that are listed as `Not assigned` on https://www.iana.org/domains/root/db, and that I manually copy-pasted by mistake.

boesing:
I am not sure if removing TLDs which were somewhen valid is a good idea. (I am aware that the script syncs TLDs from IANA).
That looks like a BC break to me.

Even tho, this is not a bugfix but QA. 

What was the previous release flow of such an update @froschdesign?

Maybe we should have some kind of `Hostname` validator which can be configured like "verify TLD is valid" aswell as "verify TLD was valid". 

If this is not considered a BC break, please remove the label. Just want to ensure precautions before releasing this.

PowerKiKi:
> not sure if removing TLDs which were somewhen valid is a good idea

I might very well be wrong, but I understand that "Not assigned" means that the TLD cannot be used by anybody to do anything at all. So the hostname is indeed invalid and cannot be used (anymore).

As a matter of fact the [previous list update](https://github.com/laminas/laminas-validator/commit/345a80a2f6cf3f2ffc2991b54cbef1737d4c7569) also removed some TLD, and it was [released in the patch release](https://github.com/laminas/laminas-validator/compare/2.13.3...2.13.4). As did the [one before](https://github.com/laminas/laminas-validator/commit/98c1923749f71f24758e108cfe4dfd2fd2c734dc). The one before, who also removed TLD was [a minor release](https://github.com/laminas/laminas-validator/commit/89a6646e0037dc2c72e5bc6b7071f03e4a0acc16). And [so on](https://github.com/laminas/laminas-validator/commit/a87f453a09fe65b0393c9f8d243140300c1a363a), [and on](https://github.com/laminas/laminas-validator/commit/f2eef40ea9b48d6175491efc77842899a29c3a6b), [and on](https://github.com/laminas/laminas-validator/commit/b594e035b7f71af88956019c4263f20afebd8be4), [and on](https://github.com/laminas/laminas-validator/commit/b3ebdf2b74d6f20a5cccf43331e1f73ae22581c6).

I'd say there is a strong case to remove TLD and release that as a patch version.

> Even tho, this is not a bugfix but QA.

I disagree, "QA" is defined as follow when creating a PR in this repo:

>   * QA improvement (additional tests, CS fixes, etc.) that does not change code
>     behavior: master branch

Changing the list of valid TLDs will change the code behavior. A domain that was incorrectly invalid, will become correctly valid. The fact that the domain was invalid is a bug, and hence this PR is a bugfix. Moreover this commit does not touch anything beside the list, especially not tests or code style related thing.

Issue #98, opened 2021-05-18
File extension comparison treats some distinct extensions as equal


Zenexer:
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | HEAD (2.15.x)

#### Summary

Certain file extensions are considered equal by validators in some contexts.  For example, `Laminas\Validator\File\ExcludeExtension#isValid` will treat file extensions `10` and `010` as equal (`example.10` and `example.010` have the same file extension in its eyes).  This is a result of using `==` and `in_array()` without strict mode to check strings.  The following both return true in PHP:

```php
'10' == '010';
in_array('10', ['010']);
```

#### Expected behavior

`example.10` should never be considered to have the file extension `010`.

fgsl:
https://github.com/laminas/laminas-validator/pull/104

gsteel:
Closed via #355

PR #104, opened 2021-07-13
File extension comparison treats some distinct extensions as equal #98

fgsl:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

This change is necessary because it fixes a bug of file extension comparison
  - You can reproduce the bug trying to compare a file with extension '10' with another with extension '010'
  - Class ExcludeExtension must return FALSE, because extensions are different
  - However, current implementation will said that both of files have the same extension.
  - TARGET THE master BRANCH

### DIFF
--- a/src/File/ExcludeExtension.php
+++ b/src/File/ExcludeExtension.php
@@ -60,7 +60,7 @@ class ExcludeExtension extends Extension
             return true;
         } elseif (! $this->getCase()) {
             foreach ($extensions as $ext) {
-                if (strtolower($ext) == strtolower($extension)) {
+                if (strtolower($ext) === strtolower($extension)) {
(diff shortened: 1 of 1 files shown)


weierophinney:
This was addressed in #10, when applying laminas-coding-standard v2, and will be released with 2.14.5.

Zenexer:
https://github.com/laminas/laminas-validator/commit/641382417563a3995809788bf9c462f8e6963b75#diff-77e911345bc3997d7fb8bbde4b8b349812819570ab56b3831e8f689967bb42c6L63

PR #107, opened 2021-09-08
Provide PHP 8.1 support

weierophinney:
This patch provides support for PHP 8.1.

Changes affecting users:

- The PHP requirement now adds `~8.1.0` to the list of allowed constraints.
- As a replacement for zend-validator:
  - The "replace" section of the package was renamed to "conflict", and the "zend-validator" entry constraint set to "*".
  - The dependency on laminas-zendframework-bridge was removed.
- The laminas-stdlib requirement was bumped to 3.6, to pick up fixes that affect PHP 8.1.
- `ValidatorChain::count()` adds a `#[ReturnTypeWillChange]` attribute to allow it to work under PHP 8.1.
- A change to `Csrf::getTokenFromHash()` to detect a `null` `$hash` argument, and return `null` immediately, avoiding passing it to `explode()` (as PHP 8.1 deprecates `null` arguments to `explode()`).
- A change to `File\Extension::getExtension()` to detect a null "extension" option, and return null early before attempting to parse it.

Changes that do not affect users:

- Since the baseline is now run against PHP 7.3, a number of new Psalm "errors" were found before any code changes occured that were all in the category of "MixedAssignment" or similar; these were added to the Psalm baseline.
- The development dependency for laminas-config was removed, as the tests that consumed it could use vanilla `Traversable` objects instead.
- A number of development dependencies are not yet updated for PHP 8.1, and required importing shims so that tests would run. Each is inlined into the test assets, and updated to work under PHP 8.1; an autoloader that checks for a PHP 8.1 runtime conditionally loads them under that version. Shims include:
  - From laminas-db:
    - `Laminas\Db\Adapter\ParameterContainer`
    - `Laminas\Db\ResultSet\AbstractResultSet`
    - `Laminas\Db\ResultSet\ResultSet`
    - `Laminas\Db\Sql\Join`
    - `Laminas\Db\Sql\PredicateSet`
  - From laminas-session:
    - `Laminas\Session\Storage\ArrayStorage`
  - From laminas-uri
    - `Laminas\Uri\Uri`
- Conditional branching in the `File\IsImageTest` and `File\IsCompressedTest` when testing constructor option population, due to a regression found in the finfo extension under PHP 8.1.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -2264,9 +2264,6 @@
     <MixedOperand occurrences="1">
       <code>$_SERVER['REQUEST_TIME']</code>
     </MixedOperand>
-    <PossiblyInvalidArgument occurrences="1">
-      <code>testTimeoutIsMutable</code>
-    </PossiblyInvalidArgument>
   </file>
   <file src="test/Db/NoRecordExistsTest.php">
     <DeprecatedMethod occurrences="2">
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="4.8.1@f73f2299dbc59a3e6c4d66cff4605176e728ee69">
+<files psalm-version="4.10.0@916b098b008f6de4543892b1e0651c1c3b92cbfa">
   <file src="bin/update_hostname_validator.php">
     <MissingClosureParamType occurrences="2">
       <code>$domain</code>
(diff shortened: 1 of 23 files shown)


psalm-baseline.xml
```diff
@@ -2711,11 +2695,13 @@
     <InvalidCast occurrences="1">
       <code>[]</code>
     </InvalidCast>
+    <MixedArgumentTypeCoercion occurrences="1">
+      <code>array_merge($validator-&gt;getMessages())</code>
```
Ocramius, line 2699: There are a lot of errors around `$this->validator->getMessages()` added to the baseline: perhaps we're missing something obvious?

Ocramius (approved):
LGTM - the added psalm issues seem to only come from a dependency upgrade, so let's leave them alone here for now (lots of them are gone anyway)

Issue #24, opened 2019-12-31
File/Count validator does not work


weierophinney:
Validation result always false. Warning dropped

Warning: dirname() expects parameter 1 to be string, array given in /home/autowp/autowp.ru/vendor/zendframework/zend-validator/src/File/Count.php on [line 209](https://github.com/zendframework/zend-validator/blob/master/src/File/Count.php#L209)

``` html
<form action="" method="post" enctype="multipart/form-data">
    <input type="file" name="file" />
    <input type="submit" />
</form>
```

``` php
$form = new \Zend\Form\Form(null);
$form->add([
    'name' => 'file',
    'type' => 'File'
]);

$inputFilter = new \Zend\InputFilter\InputFilter();
$inputFilter->add([
    'validators' => [
        [
            'name'    => \Zend\Validator\File\Count::class,
            'options' => ['min' => 1]
        ]
    ]
], 'file');

$form->setInputFilter($inputFilter);

$form->setData($this->getRequest()->getFiles());
print $form->isValid() ? 'valid' : 'invalid'; // invalid, warning dropped
```

---

Originally posted by @autowp at https://github.com/zendframework/zend-validator/issues/114

weierophinney:
The [documentation of the validator](https://docs.zendframework.com/zend-validator/validators/file/count/) is also wrong.

---

Originally posted by @froschdesign at https://github.com/zendframework/zend-validator/issues/114#issuecomment-307575382

weierophinney:
Same issue for me, the validator is call for each file in `Zend\InputFilter\FileInput@isValid` do not get the right parameters ("value" and "context" instead of "filenames" and "files")

Version of zendframework/zend-validator : 2.10.2

---

Originally posted by @bfolliot at https://github.com/zendframework/zend-validator/issues/114#issuecomment-427369833

cbichis:
Hi,

It's this fixed?

froschdesign:
@cbichis 
Unfortunately not yet, but any help is welcome. Thanks in advance!

gsteel:
Closed via #356 - This patch drops support for legacy Laminas\File\Transfer api so it's no longer an issue, also, dirname is verifiably not called on non-strings.

PR #109, opened 2021-09-09
Supported PSR-7 in File\Count validator

icetee:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description
This PR complements previous PR https://github.com/zendframework/zend-validator/pull/251. Recommendation PSR-7 UploadedFileInterface has not been implemented in File / Count validation. At the same time, it improves the https://github.com/laminas/laminas-validator/issues/24 issue.

### DIFF
--- a/src/File/Count.php
+++ b/src/File/Count.php
@@ -195,8 +195,8 @@ class Count extends AbstractValidator
             }
         }
 
-        if ($file instanceof UploadedFileInterface) {
-            $this->files[$file->getClientFilename()] = $file->getClientFilename();
+        if ($file instanceof UploadedFileInterface && is_string($file->getClientFilename())) {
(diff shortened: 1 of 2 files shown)


Ocramius (changes requested):
I like this: good improvement.

Missing an unhappy path with an `UploadedFileInterface` though :thinking:

Ocramius:
Also: sorry for not noticing the patch before :\

icetee:
I need to change something? @Ocramius

Ocramius:
I think a `false` scenario in the validator test would be good here

icetee:
I updated the base version (2.27.x), let's continue here #158

PR #112, opened 2021-10-25
Fix: add aliases for IsCountable in ValidatorPluginManager

### LINKED ISSUE #110: IsCountable validator is not registered in ValidatorPluginManager
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 2.12

#### Current behavior

The `IsCountable` validator is not registered in `ValidatorPluginManager` and can not be created with this Plugin-Manager for example with InputFilter.

#### How to reproduce

```
use Laminas\InputFilter\Factory;

$factory = new Factory();
$inputFilter = $factory->createInputFilter([
    'password' => [
        'name'       => 'c',
        'required'   => true,
        'validators' => [
            [
                'name' => 'not_empty',
            ],
            [
                'name' => 'IsCountable', // or is_countable
            ],
        ],
    ],
]);
```

#### Expected behavior

Input Filter could be created

rieschl:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Aliases for the `IsCountable` validator were missing in the `ValidatorPluginManager` alias definitions.

fixes #110


rieschl:
meh, do we really want/need to support laminas-servicemanager v2 and legacy zend aliases?

Ocramius:
@rieschl IMO having the `::class` named one is sufficient - the legacy names are not needed.

rieschl:
> @rieschl IMO having the `::class` named one is sufficient - the legacy names are not needed.

turns out that it's the legacy servicemanager v2 alias that was missing.
should we drop servicemanager v2 support (in another PR)? I think that this was done in some other packages, already.

froschdesign:
@rieschl 
> should we drop servicemanager v2 support (in another PR)?

👍 

> I think that this was done in some other packages, already.

Correct!

rieschl:
> @rieschl
> 
> > should we drop servicemanager v2 support (in another PR)?
> 
> 👍
> 
> > I think that this was done in some other packages, already.
> 
> Correct!

So should I create a PR to drop servicemanager v2 (and Zend?) just for the sake of it, or do that directly in this PR?

Ocramius:
IMO feasible/OK to drop `laminas/laminas-servicemanager:^2` support - `^3` has been out for years now, and the migration path is viable for everyone.

rieschl:
> IMO feasible/OK to drop `laminas/laminas-servicemanager:^2` support - `^3` has been out for years now, and the migration path is viable for everyone.

Okay! Directly in this PR or should I open a new one? Dropping v2 is a minor, not a patch, right? So I'd have to rebase this PR.
And what about the Zend aliases? Keep or drop? At least the Zend FQCNs. I'd probably keep the string aliases.

rieschl:
so @Ocramius @froschdesign ... should I drop the v2 of servicemanager here or in a new PR? thanks!

froschdesign:
@rieschl 
_Sorry for the very late response!_ 😞 

Please create a separate pull request to drop the old version of laminas-servicemanager. This allows us to release the current pull request as a bug-fix.

PR #120, opened 2022-03-03
Between refactor

sjokkateer:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description
Small refactor/simplification of the constructor's conditionals.

### DIFF
--- a/src/Between.php
+++ b/src/Between.php
@@ -91,19 +91,13 @@ class Between extends AbstractValidator
             $options = $temp;
         }
 
-        if (! array_key_exists('min', $options) || ! array_key_exists('max', $options)) {
+        if (! isset($options['min'], $options['max'])) {
             throw new Exception\InvalidArgumentException("Missing option: 'min' and 'max' have to be given");
         }
 
-        if (
-            (isset($options['min']) && is_numeric($options['min']))
-            && (isset($options['max']) && is_numeric($options['max']))
-        ) {
+        if (is_numeric($options['min']) && is_numeric($options['max'])) {
             $this->numeric = true;
-        } elseif (
-            (isset($options['min']) && is_string($options['min']))
-            && (isset($options['max']) && is_string($options['max']))
-        ) {
+        } elseif (is_string($options['min']) && is_string($options['max'])) {
             $this->numeric = false;
         } else {
             throw new Exception\InvalidArgumentException(
--- a/src/Between.php
+++ b/src/Between.php
@@ -5,7 +5,6 @@ namespace Laminas\Validator;
 use Laminas\Stdlib\ArrayUtils;
 use Traversable;
 
-use function array_key_exists;
 use function array_shift;
 use function func_get_args;
 use function is_array;
--- a/test/BetweenTest.php
+++ b/test/BetweenTest.php
@@ -297,6 +297,9 @@ class BetweenTest extends TestCase
             'min-undefined' => [['min' => 0, 'foo' => 'bar']],
             'max-undefined' => [['max' => 5, 'foo' => 'bar']],
             'no-min-or-max' => [['bar' => 'foo', 'foo' => 'bar']],
+            'min-null'      => [['min' => null, 'max' => 5]],
+            'max-null'      => [['min' => 18, 'max' => null]],
+            'min-max-null'  => [['min' => null, 'max' => null]],
         ];
     }
 


src/Between.php
```diff
@@ -91,19 +90,13 @@ public function __construct($options = null)
...
-        if (! array_key_exists('min', $options) || ! array_key_exists('max', $options)) {
+        if (! isset($options['min'], $options['max'])) {
             throw new Exception\InvalidArgumentException("Missing option: 'min' and 'max' have to be given");
         }
 
-        if (
-            (isset($options['min']) && is_numeric($options['min']))
-            && (isset($options['max']) && is_numeric($options['max']))
-        ) {
+        if (is_numeric($options['min']) && is_numeric($options['max'])) {
```
boesing, line 97: ~~This would also change to numeric to true when it is an integer.
Havent fully overseen the current implementation but if numeric should only be true for strings, this should be reverted.~~

~~https://3v4l.org/JQRG9~~

Ah nvm.
I was on my mobile phone this morning and thus miss-interpreted that change. This is exactly as before... Sorry.

src/Between.php
```diff
@@ -91,19 +90,13 @@ public function __construct($options = null)
...
             $this->numeric = true;
-        } elseif (
-            (isset($options['min']) && is_string($options['min']))
-            && (isset($options['max']) && is_string($options['max']))
-        ) {
+        } elseif (is_string($options['min']) && is_string($options['max'])) {
```
glensc, line 99: I think @boesing wanted to comment on this chunk:
- https://github.com/laminas/laminas-validator/pull/120/files#r819301581

src/Between.php
glensc: You probably intended to commit this chunk instead:
- https://github.com/laminas/laminas-validator/pull/120/files#r819397790

src/Between.php
boesing: ~~Kinda. The problem is, that integer values are `numeric` as well. I am not 100% sure if `numeric` is used here to somehow cast the value to integer before actually using it since I havent dig into the code.
Maybe the `numeric` flag enables some kind of quirks for e.g. floats or whatever.~~

~~So all I want to ensure is that if we toggle the `numeric` flag to `true` now for integer values, that this might introduce a change we might not be aware of.~~

~~If we are fine with setting `numeric` flag to `true` for integer values as well, my concerns can be ignored.~~

----

~~TL:DR: before this patch, only string values toggled `numeric` to true, now integer values do as well. Not sure if that is a problem but I guess some1 can verify this before merging this.~~

Ah nvm.
I was on my mobile phone this morning and thus miss-interpreted that change. This is exactly as before... Sorry.

src/Between.php
sjokkateer: The only thing that might slightly differ in response is the replacement of `array_key_exists` with `isset`, but in my opinion it does express better the intention (hopefully you guys agree).

Ocramius: `array_key_exists()` is more expressive here, and should be preferred over `isset()`, when possible.

Since PHP 7.4, AFAIK, `array_key_exists()` is also replaced with a specialized opcode, as long as it is an absolute function reference.

sjokkateer: > `array_key_exists()` is more expressive here, and should be preferred over `isset()`, when possible.
> 
> Since PHP 7.4, AFAIK, `array_key_exists()` is also replaced with a specialized opcode, as long as it is an absolute function reference.

Would you want me to change `isset` to `array_key_exists` plus test for both if the value is not `null`?
Forgive me, I meant the `isset` that I use in line 93* such that the `isset` in later lines can be ommitted.

test/BetweenTest.php
```diff
@@ -336,4 +336,40 @@ public function testNumericValidatedAgainstStringMinAndMaxIsInvalidAndReturnsAFa
...
+            'min set, max missing'     => [['min' => 18]],
+            'min missing, max missing' => [[]],
+        ];
+    }
+
+    /** @dataProvider optionsArgumentsProvider */
+    public function testMissingMinMaxOptionsSeparateConstructorArguments(?int $min, ?int $max): void
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectErrorMessage("Missing option: 'min' and 'max' have to be given");
```
sjokkateer: @Ocramius Because of the `isset` in [L93](https://github.com/laminas/laminas-validator/pull/120/files#diff-912ea778945add3d4cf4903b4bc97a2cdf688007d9a65a8ce6d5260067398c92R93) this now throws an `InvalidArgumentException::class` with the message `"Missing option: 'min' and 'max' have to be given"`. I thought that was more suiting (as a result of the refactor), plus the following conditionals are a bit more readable as a result.

sjokkateer: NVM, I was under the impression that it was possible to pass a variable number of arguments, while psalm seems to catch this part already.

PR #122, opened 2022-03-06
[WIP] BETWEEN: Add test for constructor arguments

sjokkateer:
Signed-off-by: Remy Bos <27890746+sjokkateer@users.noreply.github.com>

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | maybe
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description
The added test seems to be in line with the implementation of `Between::__construct`. This currently fails the static analysis test. This could be fixed by adding a union `int` type hint to `$options` if that is preferred.

Please let me know what you think about the proposed. If this is the wanted behavior I will also update it with an example for the documentation.

### DIFF
--- a/test/BetweenTest.php
+++ b/test/BetweenTest.php
@@ -336,4 +336,11 @@ class BetweenTest extends TestCase
             $messages
         );
     }
+
+    public function testBetweenConstructorGivenArgumentsInsteadOfOptionsArray(): void
+    {
+        $validator = new Between(1, 4, true);
+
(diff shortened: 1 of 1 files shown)


sjokkateer:
There is also [BetweenTest::testConstructorCanAcceptInclusiveParameter](https://github.com/laminas/laminas-validator/blob/2.17.x/test/BetweenTest.php#L303) in the tests.

PR #123, opened 2022-03-08
Domains with bad MX configurations should not print warnings to page

Saeven:
|    Q          |   A
|-------------- | ------
| Documentation |no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

This change is necessary to prevent the EmailAddress validator, with deep MX checks enabled, from printing warnings to page when domain configurations are not valid.

It is possible that mx records return empty servers, which is where the validator currently prints unsightly errors to page (e.g., `host example.com`).

For example, attempting to validate jon@example.com (which is a common black-box scanner test), yields:
Screen Shot 2022-03-07 at 10 47 36 PM

Validation should be silent, the warnings otherwise yield useful information to attackers (path structure, etc.).

A test is added to validate the correction (test fails when the fix is not implemented) by trapping (and then restoring) the error handler.

Cheers.
Alex

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -4,6 +4,7 @@
          bootstrap="./vendor/autoload.php"
          beStrictAboutOutputDuringTests="true"
          beStrictAboutTestsThatDoNotTestAnything="true"
+         failOnWarning="true"
          colors="true">
     <testsuites>
         <testsuite name="laminas-validator Test Suite">
--- a/test/EmailAddressTest.php
+++ b/test/EmailAddressTest.php
@@ -18,14 +18,12 @@ use function getenv;
 use function implode;
 use function next;
 use function preg_replace;
-use function restore_error_handler;
 use function set_error_handler;
 use function sprintf;
 use function str_repeat;
 use function strstr;
 
 use const E_USER_NOTICE;
-use const E_WARNING;
 
 /**
  * @group      Laminas_Validator
@@ -934,15 +932,6 @@ class EmailAddressTest extends TestCase
             'useDeepMxCheck' => true,
         ]);
 
-        $called = false;
-        /** @psalm-suppress InvalidArgument **/
-        set_error_handler(static function () use (&$called) {
-            $called = true;
-        }, E_WARNING);
-
-        $validator->isValid('jon@example.com');
-
-        restore_error_handler();
-        $this->assertFalse($called);
+        $this->assertFalse($validator->isValid('jon@example.com'));
     }
 }
--- a/test/EmailAddressTest.php
+++ b/test/EmailAddressTest.php
@@ -935,7 +935,8 @@ class EmailAddressTest extends TestCase
         ]);
 
         $called = false;
-        set_error_handler(function ($errno, $errstr, $errfile, $errline) use (&$called) {
+        /** @psalm-suppress InvalidArgument **/
+        set_error_handler(static function () use (&$called) {
             $called = true;
         }, E_WARNING);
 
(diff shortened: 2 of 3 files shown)


test/EmailAddressTest.php
```diff
@@ -924,4 +926,23 @@ public function testCanSetDomainCheckFlag(): void
...
+    public function testWillNotCheckEmptyDeepMxChecks(): void
+    {
+        $validator = new EmailAddress([
+            'useMxCheck'     => true,
+            'useDeepMxCheck' => true,
+        ]);
+
+        $called = false;
+        /** @psalm-suppress InvalidArgument **/
+        set_error_handler(static function () use (&$called) {
```
Ocramius: Warnings are generally caught by the test suite - this sort of handler setup shouldn't be necessary.

Let's instead drop it, and assert on the validity result?

Ocramius (changes requested):

Ocramius:
@Saeven this looks like a bugfix - should we target `2.16.x` perhaps?

test/EmailAddressTest.php
Saeven: Hello @Ocramius, I hope that you are doing well!  

_There might indeed be a better way to validate the fix.  Perhaps you can offer some guidance:_

In this case, the only smell of a problem is the output itself. The validity result is not sufficient, since the assertion succeeds with an expected value without this patch.  The difference is, this patch removes the warnings.

Paraphrasing, if we reduce the test to:

        $validator = new EmailAddress([
            'useMxCheck'     => true,
            'useDeepMxCheck' => true,
        ]);
        $this->assertFalse($validator->isValid('jon@example.com'));

We find that the result is valid.  The problem is, that it prints warnings to page:

```
1) LaminasTest\Validator\EmailAddressTest::testWillNotCheckEmptyDeepMxChecks
This test printed output: 
Warning: checkdnsrr(): Host cannot be empty in /Users/Saeven/Desktop/laminas-validator/src/EmailAddress.php on line 463

Warning: checkdnsrr(): Host cannot be empty in /Users/Saeven/Desktop/laminas-validator/src/EmailAddress.php on line 464

Warning: checkdnsrr(): Host cannot be empty in /Users/Saeven/Desktop/laminas-validator/src/EmailAddress.php on line 465
```

Is there a better way, with phpunit (phpspec is my goto, I don't use phpunit day to day!), to assert that a test did not print any output while it ran?

The current proposed test, was my attempt to ensure that nothing was printed - solving the problem experienced.

Cheers!
Alex

Saeven:
Definitely defer to you, as to where you would prefer to apply it. 👍🏼

test/EmailAddressTest.php
Ocramius: Hmm, perhaps we need more strictness in `phpunit.xml.dist` then? There's `beStrictAboutTestsThatProduceOutput=` or such

Saeven: I've indeed found such a creature in their docs.  failOnWarning added to phpspec xml, can confirm that it traps the issue on an unadulterated master.  Test simplified.

phpunit.xml.dist
```diff
@@ -4,6 +4,7 @@
          bootstrap="./vendor/autoload.php"
          beStrictAboutOutputDuringTests="true"
          beStrictAboutTestsThatDoNotTestAnything="true"
+         failOnWarning="true"
```
Ocramius, line 7: Wait... how is this not the default in PHPUnit? :O

Saeven: I had the same thought 💯   With PHPSpec, it's an insta-fail.

Ocramius (approved):
LGTM, thanks @Saeven!

Saeven:
Cheers @Ocramius, always a pleasure to work with you.

Ocramius:
It's in `2.17.0` :)

PR #127, opened 2022-06-08
qa: psalm improvements

gsteel:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes
| BC Break      | no
| QA            | yes

### Description

Trying to make some improvements to `input-filter` has led me down a psalm rabbit hole…

Whilst this pull is a general reduction in the baseline and improvement for AbstactValidator and the plugin manager, it has introduced some new errors that I've baselined because I don't have the time - most of them appear to be constructor incompatibilities in the many validators - i.e. annotating each individual constructor with `@param array<string, mixed>|Traversable<string, mixed> $options`

Spotted a couple of void returns when they were declared as whatever|null

I'll do a review and point out as much as possible - Hopefully, on balance this is an improvement…


src/AbstractValidator.php
```diff
@@ -26,6 +26,20 @@
...
+ *     messageTemplates: array<string, string>,
+ *     messageVariables: array<string, mixed>,
+ *     translator: Translator\TranslatorInterface|null,
+ *     translatorTextDomain: string|null,
+ *     translatorEnabled: bool,
+ *     valueObscured: bool,
+ * }
+ * @property array<string, mixed> $options
+ * @property array<string, string> $messageTemplates
+ * @property array<string, mixed> $messageVariables
```
gsteel, line 40: declaring these magic properties has caused an increase in psalm errors that should be easily fixed in individual validators

src/AbstractValidator.php
```diff
@@ -296,29 +322,29 @@ public function __get($property)
     protected function createMessage($messageKey, $value)
     {
         if (! isset($this->abstractOptions['messageTemplates'][$messageKey])) {
-            return;
+            return null;
```
gsteel, line 323: This was a bug right?

src/AbstractValidator.php
```diff
@@ -351,11 +377,19 @@ protected function error($messageKey, $value = null)
...
             $value = $this->value;
         }
 
-        $this->abstractOptions['messages'][$messageKey] = $this->createMessage($messageKey, $value);
+        $message = $this->createMessage($messageKey, $value);
+        if (! is_string($message)) {
+            return;
+        }
+
+        $this->abstractOptions['messages'][$messageKey] = $message;
```
gsteel, line 388: Message is possibly null here, so an early return prevents `null` from being added to the list of messages - perhaps a subtle BC break?

src/AbstractValidator.php
```diff
@@ -427,7 +462,7 @@ public function setTranslator(?Translator\TranslatorInterface $translator = null
     public function getTranslator()
     {
         if (! $this->isTranslatorEnabled()) {
-            return;
+            return null;
```
gsteel, line 461: Another void return which _should_ be null

src/ValidatorChain.php
```diff
@@ -239,12 +244,13 @@ public function isValid($value, $context = null)
...
         foreach ($this->validators as $element) {
             $validator = $element['instance'];
+            assert($validator instanceof ValidatorInterface);
             if ($validator->isValid($value, $context)) {
                 continue;
             }
             $result         = false;
             $messages       = $validator->getMessages();
-            $this->messages = array_replace_recursive($this->messages, $messages);
+            $this->messages = array_replace($this->messages, $messages);
```
gsteel, line 253: messages is `array<string, string>` therefore `array_replace` is sufficient and retains the existing type for psalm

src/AbstractValidator.php
```diff
@@ -26,6 +26,20 @@
...
+ * @psalm-type AbstractOptions = array{
+ *     messages: array<string, string>,
+ *     messageTemplates: array<string, string>,
+ *     messageVariables: array<string, mixed>,
+ *     translator: Translator\TranslatorInterface|null,
+ *     translatorTextDomain: string|null,
+ *     translatorEnabled: bool,
+ *     valueObscured: bool,
+ * }
+ * @property array<string, mixed> $options
```
Ocramius, line 38: Are these `public`? Because `@property` makes them `public` :thinking:

src/AbstractValidator.php
```diff
@@ -78,20 +91,23 @@ abstract class AbstractValidator implements
...
      *
-     * @param array|Traversable $options
+     * @param array<string, mixed>|Traversable<string, mixed> $options
      */
     public function __construct($options = null)
     {
         // The abstract constructor allows no scalar values
         if ($options instanceof Traversable) {
+            /** @psalm-var array<string, mixed> $options */
             $options = ArrayUtils::iteratorToArray($options);
```
Ocramius, line 99: TBH, sounds like we could improve inference here, in future

src/AbstractValidator.php
froschdesign: Very old remains.

src/AbstractValidator.php
froschdesign: Same here.

src/AbstractValidator.php
gsteel: I'll send in a pull for `ArrayUtils::iteratorToArray` asap

src/AbstractValidator.php
gsteel: I guess if they're undeclared they are already public no? Typically they'd be protected in extending classes, if set at all, so I guess not. I'm unsure as to why `$abstractOptions` exists at all and why these properties aren't just declared on the abstract

All I actually wanted to achieve was to nail down what `getMessages()` returns but doing that is a bit of an endless battle.

src/AbstractValidator.php
gsteel: https://github.com/laminas/laminas-stdlib/pull/62

Ocramius: @gsteel released!

.psr-container-stub.php
```diff
@@ -0,0 +1,21 @@
+<?php
```
Ocramius: Following discussion at https://github.com/Lctrs/psalm-psr-container-plugin/issues/545#issuecomment-1150465946, can we:

 1. export-ignore this
 2. perhaps rename it to `.psr-container-stub.php.stub`?

Ocramius: We would probably drop a line of comment explaining why this doesn't end with `.php` :D

src/AbstractValidator.php
Ocramius: Let's keep it as-is then: still a good improvement :+1:

src/AbstractValidator.php
Ocramius: I think it's OK to prevent `null` from leaking in here.

Ocramius (approved):
LGTM! Waiting for @gsteel's feedback and potential changes on the stub file, before merging

.psr-container-stub.php
gsteel: 👍 - Had a general tidy-up of the dot-files/ignores etc too and got rid of coveralls

Ocramius:
Thanks @gsteel!

.psr-container-stub.php
Ocramius: Indeed, awesome!

PR #128, opened 2022-06-09
Adds type inference for `plugin()` when given a class-string

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Updates `ValidatorChain::plugin()` so that given a validator class-string there's no need to assert the return type


Ocramius (approved):
Thanks @gsteel

PR #144, opened 2022-07-27
Apply PHP 7.4 syntax and typed property

samsonasik:
Signed-off-by: Abdul Malik Ikhsan <samsonasik@gmail.com>

|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Since composer.json require php 7.4, php 7.4 syntax can be applied with typed properties. 

- For `final class`, typed properties for protected is allowed if no inherit of parent or used by its final class parent.
- Otherwise, update only private properties.

### DIFF
--- a/src/UndisclosedPassword.php
+++ b/src/UndisclosedPassword.php
@@ -39,13 +39,9 @@ final class UndisclosedPassword extends AbstractValidator
         self::NOT_A_STRING      => 'The provided password is not a string, please provide a correct password',
     ];
 
-    // phpcs:enable
+    private ClientInterface $httpClient;
 
-    /** @var ClientInterface */
-    private $httpClient;
-
-    /** @var RequestFactoryInterface */
-    private $makeHttpRequest;
+    private RequestFactoryInterface $makeHttpRequest;
 
     public function __construct(ClientInterface $httpClient, RequestFactoryInterface $makeHttpRequest)
     {
--- a/test/CallbackTest.php
+++ b/test/CallbackTest.php
@@ -90,9 +90,7 @@ class CallbackTest extends TestCase
     {
         $value     = 'bar';
         $context   = ['foo' => 'bar', 'bar' => 'baz'];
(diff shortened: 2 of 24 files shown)


samsonasik:
All green 🎉

src/UndisclosedPassword.php
```diff
@@ -39,13 +39,9 @@ final class UndisclosedPassword extends AbstractValidator
         self::NOT_A_STRING      => 'The provided password is not a string, please provide a correct password',
     ];
 
-    // phpcs:enable
-
-    /** @var ClientInterface */
-    private $httpClient;
+    private ClientInterface $httpClient;
```
andrew-demb, line 42:
```suggestion
    // phpcs:enable

    private ClientInterface $httpClient;
```

Ocramius (approved):
Thanks @samsonasik !

src/UndisclosedPassword.php
Ocramius: Ha, CS completely disabled here, didn't notice :O

test/CallbackTest.php
```diff
@@ -90,9 +90,7 @@ public function testCanAcceptContextWithoutOptions(): void
     {
         $value     = 'bar';
         $context   = ['foo' => 'bar', 'bar' => 'baz'];
-        $validator = new Callback(function ($v, $c) use ($value, $context) {
-            return ($value === $v) && ($context === $c);
-        });
+        $validator = new Callback(static fn($v, $c) => ($value === $v) && ($context === $c));
```
holtkamp, line 93: Why not use a return type ` : bool` here?

Ocramius: @holtkamp I think that requires more additions to the setup that @samsonasik uses (he's a [Rector](https://github.com/rectorphp/rector) maintainer)

samsonasik: Yes, that's can be handled by type declaration set list, it was missed cherry pick return type declaration rule to apply, though need to only apply to ArrowFunction

PR #149, opened 2022-09-12
Fix some 8.2 compatibility issues in `Date` validator

gsteel:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes
| QA            | yes

### Description

Testing a local project under 8.2-RC reveals a problem specific to the Date validator, where `DateTime::getLastErrors()` is potentially false. Even given https://3v4l.org/HBq0q - I can't get a test to actually return false from this method because it's 8.2 specific, so verifying the fix is a tough one.

Looking at https://3v4l.org/3QsKY - Only 8.2 returns false *after* a date operation that causes no errors.

A couple of other tests had complaints about dynamic properties and I've silenced a number of tests that don't perform any assertions depending on the data providers.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="4.24.0@06dd975cb55d36af80f242561738f16c5f58264f">
+<files psalm-version="4.26.0@6998fabb2bf528b65777bf9941920888d23c03ac">
   <file src="bin/update_hostname_validator.php">
     <MissingClosureParamType occurrences="2">
       <code>$domain</code>
@@ -579,6 +579,9 @@
     <PossiblyUndefinedVariable occurrences="1">
       <code>$temp</code>
     </PossiblyUndefinedVariable>
+    <TypeDoesNotContainType occurrences="1">
+      <code>$errors === false</code>
+    </TypeDoesNotContainType>
   </file>
   <file src="src/DateStep.php">
     <ArgumentTypeCoercion occurrences="1">
@@ -1431,7 +1434,7 @@
     </PossiblyNullArgument>
     <RedundantConditionGivenDocblockType occurrences="3">
       <code>$file !== null</code>
-      <code>is_array($files)</code>
+      <code>is_countable($files)</code>
     </RedundantConditionGivenDocblockType>
   </file>
   <file src="src/File/UploadFile.php">
@@ -2045,9 +2048,6 @@
       <code>[$this, 'errorHandlerIgnore']</code>
       <code>[$this, 'errorHandlerIgnore']</code>
     </InvalidArgument>
-    <InvalidScalarArgument occurrences="1">
-      <code>1</code>
-    </InvalidScalarArgument>
     <MixedArgument occurrences="2">
       <code>$message</code>
       <code>$options</code>
@@ -2059,9 +2059,6 @@
       <code>null</code>
       <code>null</code>
     </NullArgument>
-    <PossiblyNullArgument occurrences="1">
-      <code>var_export($messages, 1)</code>
-    </PossiblyNullArgument>
   </file>
   <file src="test/BarcodeTest.php">
     <ArgumentTypeCoercion occurrences="1">
@@ -2167,9 +2164,6 @@
       <code>'users'</code>
       <code>'users'</code>
     </InvalidArgument>
-    <UndefinedPropertyAssignment occurrences="1">
-      <code>$mockHasResultRow-&gt;one</code>
-    </UndefinedPropertyAssignment>
   </file>
   <file src="test/Db/RecordExistsTest.php">
     <DeprecatedMethod occurrences="2">
@@ -2188,9 +2182,6 @@
     <PossiblyNullReference occurrences="1">
       <code>$parameters</code>
     </PossiblyNullReference>
-    <UndefinedPropertyAssignment occurrences="1">
-      <code>$mockHasResultRow-&gt;one</code>
-    </UndefinedPropertyAssignment>
   </file>
   <file src="test/Db/TestAsset/ConcreteDbValidator.php">
     <PropertyNotSetInConstructor occurrences="2">
@@ -2684,31 +2675,11 @@
     <ArgumentTypeCoercion occurrences="1">
       <code>$classOrInstance</code>
     </ArgumentTypeCoercion>
-    <InvalidPropertyAssignmentValue occurrences="1"/>
     <MixedArgument occurrences="2"/>
     <MixedInferredReturnType occurrences="2">
       <code>array</code>
       <code>array</code>
     </MixedInferredReturnType>
-    <MixedMethodCall occurrences="6">
-      <code>will</code>
-      <code>will</code>
-      <code>willReturn</code>
-      <code>willReturn</code>
-      <code>willReturnCallback</code>
-      <code>willReturnCallback</code>
-    </MixedMethodCall>
-    <PossiblyNullPropertyAssignmentValue occurrences="1">
-      <code>null</code>
-    </PossiblyNullPropertyAssignmentValue>
-    <UndefinedInterfaceMethod occurrences="6">
-      <code>method</code>
-      <code>method</code>
-      <code>method</code>
-      <code>method</code>
-      <code>method</code>
-      <code>method</code>
-    </UndefinedInterfaceMethod>
   </file>
   <file src="test/UriTest.php">
     <InvalidArgument occurrences="4">
--- a/test/File/Crc32Test.php
+++ b/test/File/Crc32Test.php
@@ -98,14 +98,14 @@ class Crc32Test extends TestCase
      */
     public function testLegacy($options, $isValidParam, bool $expected, string $messageKey): void
     {
-        if (is_array($isValidParam)) {
-            $validator = new File\Crc32($options);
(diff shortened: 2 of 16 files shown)


psalm-baseline.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="4.24.0@06dd975cb55d36af80f242561738f16c5f58264f">
+<files psalm-version="4.26.0@6998fabb2bf528b65777bf9941920888d23c03ac">
```
boesing, line 2: cant we already bump to v4.27?

test/File/ExcludeExtensionTest.php
```diff
@@ -100,6 +100,8 @@ public function testLegacy($options, $isValidParam, bool $expected, string $mess
             if (! $expected) {
                 $this->assertArrayHasKey($messageKey, $validator->getMessages());
             }
+        } else {
+            $this->expectNotToPerformAssertions();
```
boesing: Can we probably change the check above to ensure test is only executed incase of an array and if its not an error, mark it as skipped? 

Feels weird if we assert there are no assertions if there is actually nothing tested.

Skipped message could be something like `only array foo is being tested for legacy whatever`.

test/File/ExistsTest.php
```diff
@@ -78,6 +78,8 @@ public function testLegacy(string $options, $isValidParam, bool $expected): void
         if (is_array($isValidParam)) {
             $validator = new File\Exists($options);
             $this->assertEquals($expected, $validator->isValid($isValidParam['tmp_name'], $isValidParam));
+        } else {
+            $this->expectNotToPerformAssertions();
```
boesing: Can we probably change the check above to ensure test is only executed incase of an array and if its not an error, mark it as skipped? 

Feels weird if we assert there are no assertions if there is actually nothing tested.

Skipped message could be something like `only array foo is being tested for legacy whatever`.

test/File/ExtensionTest.php
```diff
@@ -101,6 +101,8 @@ public function testLegacy($options, $isValidParam, bool $expected, string $mess
             if (! $expected) {
                 $this->assertArrayHasKey($messageKey, $validator->getMessages());
             }
+        } else {
+            $this->expectNotToPerformAssertions();
```
boesing: Can we probably change the check above to ensure test is only executed incase of an array and if its not an error, mark it as skipped? 

Feels weird if we assert there are no assertions if there is actually nothing tested.

Skipped message could be something like `only array foo is being tested for legacy whatever`.

test/File/HashTest.php
```diff
@@ -132,6 +132,8 @@ public function testLegacy($options, $isValidParam, bool $expected, string $mess
             if (! $expected) {
                 $this->assertArrayHasKey($messageKey, $validator->getMessages());
             }
+        } else {
+            $this->expectNotToPerformAssertions();
```
boesing: Can we probably change the check above to ensure test is only executed incase of an array and if its not an error, mark it as skipped? 

Feels weird if we assert there are no assertions if there is actually nothing tested.

Skipped message could be something like `only array foo is being tested for legacy whatever`.

test/File/ImageSizeTest.php
```diff
@@ -168,6 +168,8 @@ public function testLegacy(array $options, $isValidParam, bool $expected, $messa
                     $this->assertArrayHasKey($messageKey, $validator->getMessages());
                 }
             }
+        } else {
+            $this->expectNotToPerformAssertions();
```
boesing: Can we probably change the check above to ensure test is only executed incase of an array and if its not an error, mark it as skipped? 

Feels weird if we assert there are no assertions if there is actually nothing tested.

Skipped message could be something like `only array foo is being tested for legacy whatever`.

test/File/Md5Test.php
```diff
@@ -126,6 +126,8 @@ public function testLegacy($options, $isValidParam, bool $expected, string $mess
             if (! $expected) {
                 $this->assertArrayHasKey($messageKey, $validator->getMessages());
             }
+        } else {
+            $this->expectNotToPerformAssertions();
```
boesing: Can we probably change the check above to ensure test is only executed incase of an array and if its not an error, mark it as skipped? 

Feels weird if we assert there are no assertions if there is actually nothing tested.

Skipped message could be something like `only array foo is being tested for legacy whatever`.

test/File/NotExistsTest.php
```diff
@@ -79,6 +79,8 @@ public function testLegacy(string $options, $isValidParam, bool $expected): void
         if (is_array($isValidParam)) {
             $validator = new File\NotExists($options);
             $this->assertEquals($expected, $validator->isValid($isValidParam['tmp_name'], $isValidParam));
+        } else {
+            $this->expectNotToPerformAssertions();
```
boesing: Can we probably change the check above to ensure test is only executed incase of an array and if its not an error, mark it as skipped? 

Feels weird if we assert there are no assertions if there is actually nothing tested.

Skipped message could be something like `only array foo is being tested for legacy whatever`.

test/File/Sha1Test.php
```diff
@@ -109,6 +109,8 @@ public function testLegacy($options, $isValidParam, bool $expected, string $mess
             if (! $expected) {
                 $this->assertArrayHasKey($messageKey, $validator->getMessages());
             }
+        } else {
+            $this->expectNotToPerformAssertions();
```
boesing: Can we probably change the check above to ensure test is only executed incase of an array and if its not an error, mark it as skipped? 

Feels weird if we assert there are no assertions if there is actually nothing tested.

Skipped message could be something like `only array foo is being tested for legacy whatever`.

test/File/SizeTest.php
```diff
@@ -87,6 +87,8 @@ public function testLegacy($options, $isValidParam, bool $expected): void
         if (is_array($isValidParam)) {
             $validator = new File\Size($options);
             $this->assertEquals($expected, $validator->isValid($isValidParam['tmp_name'], $isValidParam));
+        } else {
+            $this->expectNotToPerformAssertions();
```
boesing: Can we probably change the check above to ensure test is only executed incase of an array and if its not an error, mark it as skipped? 

Feels weird if we assert there are no assertions if there is actually nothing tested.

Skipped message could be something like `only array foo is being tested for legacy whatever`.

test/File/WordCountTest.php
```diff
@@ -82,6 +82,8 @@ public function testLegacy($options, $isValidParam, bool $expected): void
         if (is_array($isValidParam)) {
             $validator = new File\WordCount($options);
             $this->assertEquals($expected, $validator->isValid($isValidParam['tmp_name'], $isValidParam));
+        } else {
+            $this->expectNotToPerformAssertions();
```
boesing: Can we probably change the check above to ensure test is only executed incase of an array and if its not an error, mark it as skipped? 

Feels weird if we assert there are no assertions if there is actually nothing tested.

Skipped message could be something like `only array foo is being tested for legacy whatever`.

test/File/Crc32Test.php
```diff
@@ -104,6 +104,8 @@ public function testLegacy($options, $isValidParam, bool $expected, string $mess
             if (! $expected) {
                 $this->assertArrayHasKey($messageKey, $validator->getMessages());
             }
+        } else {
+            $this->expectNotToPerformAssertions();
```
boesing: Can we probably change the check above to ensure test is only executed incase of an array and if its not an error, mark it as skipped? 

Feels weird if we assert there are no assertions if there is actually nothing tested.

Skipped message could be something like `only array foo is being tested for legacy whatever`.

boesing (changes requested):
Does psalm already support that `false` return value of `DateTime#getLastErrors`? 

If so, its absolutely okay to early return without actually having a test which verifies if the methods return value is `false` is due to the fact that `DateTime::getLastErrors` returns an array with `warning_count`/`error_count` being positive-int  or due to the return value `false`.

Overall, LGTM.

gsteel:
Related: https://github.com/vimeo/psalm/pull/8478

psalm-baseline.xml
gsteel: I can do that separately - I didn't want the noise in this pull

test/File/ExcludeExtensionTest.php
gsteel: Yep - that was laziness on my part. All relevant tests adjusted to skip non-array args.

boesing:
@gsteel you actually have to disable `ignoreInternalFunctionFalseReturn` in psalm config.

Sadly, thats enabled by default.

Ref: https://github.com/vimeo/psalm/pull/7758#issuecomment-1228577723

gsteel:
@boesing - this should be good to go. The conditional in `Date` validator is required for PHP 8.2 though due to real errors such as `array offset on bool`… so I guess we'll just leave it in the baseline.

I did take a quick look at disabling `ignoreInternalFunctionFalseReturn` in `psalm.xml` but that yielded another 200 errors. Happy to disable it and expand the baseline if you'd like to see that here?

Is it worth trawling through the options and coming up with a standard config for all of the components do you think?

Ocramius:
> Is it worth trawling through the options and coming up with a standard config for all of the components do you think?

Yes, but not here :D

Ocramius (approved):
LGTM, thanks @gsteel!

PR #152, opened 2022-09-20
Mark the ValidatorChain as iterable and final

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes/no
| Bugfix        | yes/no
| BC Break      | yes/no
| New Feature   | yes

### Description

This patch has less value than https://github.com/laminas/laminas-filter/pull/64 - initial investigation shows that other components consistently call `getValidators()` when they wish to iterate over the list of validators - the reason for this is that each element the chain returns is an array of a specific shape: `{instance: ValidatorInterface, break_chain_on_failure: bool}`

Feel free to close if it is simply not a worthwhile change :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
     "require": {
         "php": "^7.4 || ~8.0.0 || ~8.1.0",
         "laminas/laminas-servicemanager": "^3.12.0",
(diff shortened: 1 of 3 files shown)


Ocramius (approved):
LGTM 👍

Issue #49, opened 2020-01-15
Validator result interface proposal


michalbundyra:
This is a proposal of ResultInterface about the stateless validator proposal made in:
- https://github.com/zendframework/zend-validator/issues/1
- https://github.com/bakura10/zend-validator/pull/1

/cc @bakura10 @zendframework/community-review-team 

---

Originally posted by @Maks3w at https://github.com/zendframework/zend-validator/pull/24

michalbundyra:
I agree with @manuakasam , this is highly confusing. Usage in my PR is much easier. People should not having to deal with two different objects for retrieving messages error. The workflow should staty like this:

``` php
$result = $validator->validate('foo');

if (!$result->isValid()) {
   foreach ($result->getMessages() as $message) {
      echo $message;
   }
}
```

This is the typical and most common use case. Translation should only be used for people that needs it. I really don't like this proposal :(.


---

Originally posted by @bakura10 at https://github.com/zendframework/zend-validator/pull/24#issuecomment-123032528

michalbundyra:
Also, I'm not sure what you are trying to do here. Trying to integrate the validation result stuff into the 2.x branch? This is a BC, all the validators are not stateless, you are going to go through a lot of pain and hazardous process if you try to make this BC with current architecture.

Also, there are other things I'd like to refactor in the validators themselves to make them more coherent with ZF conventions, so I'd prefer to do all the BC at once.


---

Originally posted by @bakura10 at https://github.com/zendframework/zend-validator/pull/24#issuecomment-123033096

michalbundyra:
Also please have a look at my own implementation for different way to solve the same issue https://github.com/bakura10/zend-validator/pull/1


---

Originally posted by @bakura10 at https://github.com/zendframework/zend-validator/pull/24#issuecomment-123162846

michalbundyra:
I like the idea of @weierophinney about to use a helper function for to translate the message 

I would like to extend de concept of translate to not only an **optional** feature for `HumanLanguageA <=> HumanLanguageB` but make it a **requirement** for do `ErrorCode => HumanLanguageX`

In such case the `translation tool` should look for the `error code` in a `message collection` and interpolate the message with the `validation context` and `values cardinality`

In resume:
- Message templates associated to a validator (text-domain)
- Validation error code for choose a single template (translation template)
- A validation context collection of values for freedom on make contextualized error messages.
- Each value have a cardinality associated for proper use of plurals, etc.


---

[...]

PR #153, opened 2022-09-21
[RFC] Stateless Validators (Again)

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| New Feature   | yes
| RFC           | yes

### Description

Is there any appetite for refactoring the validator component to return a validation result, roughly as described in this PR?

The theory here is that error messages/validation failures are carried around by the immutable validation result and translation of the messages would be performed on/by the result using a composed translator (if any).

This would allow validators to behave more like stateless services.

A refactoring like this has been discussed in the past: #48, #38, #49


gsteel:
Obviously, I've made absolutely no attempt to make CI pass here!

src/StringLength.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
```
Ocramius, line 3: IMO too risky for a minor release

src/ValidationResult.php
```diff
@@ -0,0 +1,217 @@
...
+use function is_resource;
+use function is_string;
+use function method_exists;
+use function sprintf;
+use function str_repeat;
+use function str_replace;
+use function strlen;
+
+/** @implements IteratorAggregate<string, non-empty-string> */
+final class ValidationResult implements IteratorAggregate, Countable
```
Ocramius: I don't think `ValidationResult` should be iterable nor countable: from a consumer PoV, it should only be iterable+countable if it failed validation

src/StringLength.php
```diff
@@ -13,32 +15,30 @@
...
-    ];
-
-    /** @var array<string, mixed> */
-    protected $options = [
+    /** @var OptionsArray */
+    private array $options = [
         'min'      => 0, // Minimum length
         'max'      => null, // Maximum length, null if there is no length limitation
         'encoding' => 'UTF-8', // Encoding to use
-        'length'   => 0, // Actual length
```
Ocramius, line 41: Possible breakage?

src/StringLength.php
```diff
@@ -13,32 +15,30 @@
 use function is_string;
 use function max;
 
-class StringLength extends AbstractValidator
```
Ocramius, line 16: Removal of ` extends AbstractValidator` is a breakage

src/ValidationResult.php
```diff
@@ -0,0 +1,217 @@
...
+    public static function new(
+        array $templates,
+        array $variables,
+        mixed $value,
+        bool $valueObscured = false,
+        ?TranslatorInterface $translator = null,
+        string $textDomain = 'default',
+    ): self {
+        return new self($templates, $variables, $value, $valueObscured, $translator, $textDomain);
+    }
```
Ocramius: We should probably have two separate named constructors:

 1. failure
 2. success

See http://marcosh.github.io/post/2017/06/16/maybe-in-php.html and http://marcosh.github.io/post/2017/10/27/maybe-in-php-2.html

src/ValidationResult.php
```diff
@@ -0,0 +1,217 @@
...
+use function is_object;
+use function is_resource;
+use function is_string;
+use function method_exists;
+use function sprintf;
+use function str_repeat;
+use function str_replace;
+use function strlen;
+
+/** @implements IteratorAggregate<string, non-empty-string> */
```
Ocramius: Should be `@psalm-immutable`

src/ValidationResult.php
```diff
@@ -0,0 +1,217 @@
...
+        }
+
+        if (is_resource($value)) {
+            return 'resource';
+        }
+
+        assert(is_object($value));
+
+        return self::stringifyObject($value);
+    }
```
Ocramius: Not sure if it's the validator's responsibility, or the view helper's :thinking:

Ocramius: Not a major issue for me though.

src/ValidationResult.php
gsteel: So do you think it would be better to define different types for success and failure states?

Ocramius: Yes, it would help:

 * while debugging
 * to get better types for us internally
 * to make it crash in case somebody tries to read error messages out of a success scenario

src/ValidationResult.php
gsteel: Mostly, this was about being absolutely rigid about the expected value of the list of error messages, i.e. `array<string, non-empty-string>` and because I put translation inside the result, I had to implement some kind of casting for the possible substitutions. It would be cleaner to make the validators provide strings to the constructor, but that would bloat all the validators wouldn't it?

If translation was external to both validators _and_ the result, i.e. `$validatorTranslator->translate($validationResult)` it would be even cleaner, but more responsibility is shifted to the view, i.e. where do they get the validator translator from? Also, the result would need to expose everything the translator needs, the message templates, keys, value, substitution keys and values.

src/ValidationResult.php
gsteel: The initial idea here is that it has a kind of builder interface like (inside a validator):

```php
$result = ValidationResult::new($myTemplates, $substitutions, $value, false, $translator);

if (! is_string($value)) {
   return $result->withError($someKey);
}

if (! $this->constraintA($value)) {
   $result = $result->withError($otherKey);
}

if (! $this->constraintB($value)) {
   $result = $result->withError($anotherKey);
}

return $result;
```

froschdesign:
@gsteel 
See also: https://github.com/weierophinney/zend-datavalidator

src/StringLength.php
```diff
@@ -13,32 +15,30 @@
...
+/**
+ * @psalm-type OptionsArray array{
+ *     min: int,
```
internalsystemerror, line 20: `int<0, max>` to prevent negative numbers (although that would asserting)

src/StringLength.php
```diff
@@ -13,32 +15,30 @@
...
+ *     max: int|null,
```
internalsystemerror, line 21: `int<0, max>|null` to prevent negative numbers (although that would asserting)

gsteel:
Closing this here - Stateless validators will not be a thing in 3.0 - more likely to be feasible in 4.0

PR #158, opened 2022-10-16
Supported PSR-7 in File\Count validator

icetee:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | yes
| New Feature   | no
| RFC           | no
| QA            | no

### Description
This PR complements previous PR https://github.com/zendframework/zend-validator/pull/251. Recommendation PSR-7 UploadedFileInterface has not been implemented in File / Count validation. At the same time, it improves the https://github.com/laminas/laminas-validator/issues/24 issue.

This PR is same https://github.com/laminas/laminas-validator/pull/109, but upgrade Count path for Actions workflow tests.

### DIFF
--- a/src/File/Count.php
+++ b/src/File/Count.php
@@ -4,6 +4,7 @@ namespace Laminas\Validator\File;
 
 use Laminas\Validator\AbstractValidator;
 use Laminas\Validator\Exception;
+use Psr\Http\Message\UploadedFileInterface;
 use Traversable;
 
 use function array_key_exists;
@@ -177,7 +178,7 @@ class Count extends AbstractValidator
     /**
      * Adds a file for validation
      *
-     * @param string|array $file
+     * @param string|array|UploadedFileInterface $file
      * @return $this
      */
     public function addFile($file)
@@ -194,6 +195,10 @@ class Count extends AbstractValidator
             }
         }
 
+        if ($file instanceof UploadedFileInterface && is_string($file->getClientFilename())) {
+            $this->files[(string) $file->getClientFilename()] = $file->getClientFilename();
+        }
+
         return $this;
     }
 
@@ -202,17 +207,23 @@ class Count extends AbstractValidator
      * not bigger than max (when max is not null). Attention: When checking with set min you
      * must give all files with the first call, otherwise you will get a false.
      *
-     * @param  string|array $value Filenames to check for count
-     * @param  array        $file  File data from \Laminas\File\Transfer\Transfer
+     * @param  string|array|UploadedFileInterface $value Filenames to check for count
+     * @param  array                              $file  File data from \Laminas\File\Transfer\Transfer
      * @return bool
      */
     public function isValid($value, $file = null)
     {
-        if (($file !== null) && ! array_key_exists('destination', $file)) {
+        $isUploadedFileInterface = $value instanceof UploadedFileInterface;
+
+        if ($isUploadedFileInterface) {
+            $this->addFile($value);
+        }
+
+        if (! $isUploadedFileInterface && ($file !== null) && ! array_key_exists('destination', $file)) {
             $file['destination'] = dirname($value);
         }
 
-        if (($file !== null) && array_key_exists('tmp_name', $file)) {
+        if (! $isUploadedFileInterface && ($file !== null) && array_key_exists('tmp_name', $file)) {
             $value = $file['destination'] . DIRECTORY_SEPARATOR . $file['name'];
         }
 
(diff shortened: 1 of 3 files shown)


Ocramius (approved):
LGTM, asking for an additional review though.

test/File/CountTest.php
```diff
@@ -241,4 +244,23 @@ public function testCanProvideMinAndMaxAsDiscreteConstructorArguments(): void
...
+    public function testPsr7FileTypes(): void
+    {
+        $testFile = __DIR__ . '/_files/testsize.mo';
+
+        $upload = $this->createMock(UploadedFileInterface::class);
+        $upload->method('getClientFilename')->willReturn(basename($testFile));
+
+        $validValidator = new Count(['min' => 1]);
+
+        $this->assertSame(true, $validValidator->isValid($upload));
```
gsteel: Style nitpick - `assertTrue` and `assertFalse` make more sense here.

gsteel (approved):
LGTM 👍 - Updating the assertions would be nice, but not a blocker IMO

Ocramius:
I just noted that this package does not depend on `psr/http-message`, which is only in `require-dev`:

https://github.com/laminas/laminas-validator/blob/23e27edc1520c0bf0c9c7e7f192dd1314677969e/composer.json#L50

I think declaring the dependency, although annoying, would be wise here.

gsteel:
> I just noted that this package does not depend on `psr/http-message`, which is only in `require-dev`:
> 
> https://github.com/laminas/laminas-validator/blob/23e27edc1520c0bf0c9c7e7f192dd1314677969e/composer.json#L50
> 
> I think declaring the dependency, although annoying, would be wise here.

I was going to suggest that too, but it is already listed in `suggest` - for the same reasoning we should require everything else too?? I would prefer it that way FWIW, but `instanceof` has got our back here right?

Ocramius:
> but `instanceof` has got our back here right?

Not if a new major version breaks BC there :|

gsteel:
> Not if a new major version breaks BC there :|

Good point - tests wouldn't catch that because dev is pinned to ^1

src/File/Count.php
```diff
@@ -202,17 +207,23 @@ public function addFile($file)
...
+        if ($isUploadedFileInterface) {
+            $this->addFile($value);
+        }
+
+        if (! $isUploadedFileInterface && ($file !== null) && ! array_key_exists('destination', $file)) {
             $file['destination'] = dirname($value);
         }
 
-        if (($file !== null) && array_key_exists('tmp_name', $file)) {
+        if (! $isUploadedFileInterface && ($file !== null) && array_key_exists('tmp_name', $file)) {
```
internalsystemerror: You could combine these ifs to reduce the amount of checks (e,g,):
```php
if ($isUploadedFileInterface) {
    $this->addFile($value);
} elseif ($file !== null) {
    if (!array_key_exists('destination', $file)) {
        $file['destination'] = dirname($value);
    }
    if (array_key_exists('tmp_name', $file)) {
        $value = $file['destination'] . DIRECTORY_SEPARATOR . $file['name'];
    }
}
```

And if you don't want the else just move it to private method with an early return.

icetee:
I corrected what came up.

UploadedFileInterface is used in several places in non-test code.
https://github.com/laminas/laminas-validator/blob/a995b21d18c63cd1f5d123d0d2cd31a1c2d828dc/src/File/Upload.php#L91
https://github.com/laminas/laminas-validator/blob/a995b21d18c63cd1f5d123d0d2cd31a1c2d828dc/src/File/Upload.php#L134
https://github.com/laminas/laminas-validator/blob/a995b21d18c63cd1f5d123d0d2cd31a1c2d828dc/src/File/Upload.php#L187
https://github.com/laminas/laminas-validator/blob/a995b21d18c63cd1f5d123d0d2cd31a1c2d828dc/src/File/Upload.php#L193
https://github.com/laminas/laminas-validator/blob/a995b21d18c63cd1f5d123d0d2cd31a1c2d828dc/src/File/Upload.php#L262

internalsystemerror (approved):
LGTM - However, I'll let @Ocramius address his dependency/BC concerns prior to merge.

Ocramius:
Yeah, for me, PSR-7 needs to become a hard dependency here...

Ocramius:
I'd let @gsteel call the shots here: meanwhile moving to `2.28.x`.

gsteel:
Thanks @icetee 👍

PR #160, opened 2022-10-31
Allowed check multiple Hostname/Uri in InputFilter

icetee:
Signed-off-by: Tamás András Horváth <htomy92@gmail.com>

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

I had to put together a field in a form that could contain several URLs and I had to validate it. I was able to make it count, but the URI validation didn't work because its input parameter could only be a string.

```php
$this->add([
    'name'        => 'links',
    'validators'  => [
        new Validator\IsCountable([
            'min'      => 0,
            'max'      => 5,
        ]),
        new Validator\Uri([
            'allowRelative' => false,
            'allowAbsolute' => true,
        ]),
    ],
    'filters'     => [
        new Filter\StringTrim(),
    ],
]);
```

### DIFF
--- a/src/Hostname.php
+++ b/src/Hostname.php
@@ -55,7 +55,7 @@ class Hostname extends AbstractValidator
     /** @var array */
     protected $messageTemplates = [
         self::CANNOT_DECODE_PUNYCODE  => "The input appears to be a DNS hostname but the given punycode notation cannot be decoded",
-        self::INVALID                 => "Invalid type given. String expected",
+        self::INVALID                 => "Invalid type given. String or Array expected",
         self::INVALID_DASH            => "The input appears to be a DNS hostname but contains a dash in an invalid position",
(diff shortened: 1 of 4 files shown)


Ocramius (changes requested):
Accepting a `list<string>` here is not a good idea: please use the validator on each array value individually instead.

You can use a callback validator, or you can use a fieldset.

Ocramius:
Note: the reason why this isn't a good idea is that:

 * some validators operate on sets with a cardinality of 1 (most scalar validators
 * some validators operate on sets with a cardinality of 0 or 1 (existence checks)

To create a validator over a set with cardinality N, you can either:

 * create N validators (what a fieldset does)
 * apply the same validator on N items in the set

Issue #167, opened 2022-12-29
Enum-Validator 


reinfi:
### Feature Request

I would like to have an enum validator. 

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | may be, it depends

#### Summary

Like I have a value and if it is in the specific enum it is valid. 

Of course this would be a php > 8.0 feature. So I just want to know if you're open for that or if it is not in the scope of this package. 

I you're open for it I will try to add a pull request :-)

Ocramius:
Perhaps just an expansion of the `InArray` validator for PHP > 8.1?

reinfi:
Like adding an Enum Class as Haystack? 

Would make it a bit complicated to differentiate and you need to mix types.

Ocramius:
Maybe: let's start with a proposal patch though, so go for it 💪

Please test it 😁

reinfi:
`public function setHaystack(array $haystack)`

just to be sure, I will need to change that to something supporting array and enum values and need to add a PHP-Version check because there is still 8.0 support.

PR #168, opened 2022-12-29
add handling for enum as haystack

reinfi:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

See issue #167.

Suggestion was to allow the haystack to be an enum. Otherwise a new validator for just the case of enums would be suitable.

I am a bit unsure about the testcase and I hope it would work with the build pipeline, but I will see it afterwards. 

And psalm is not working because it misses the classes on PHP 8.0.

Happy to hear your feedback.

### DIFF
--- a/src/InArray.php
+++ b/src/InArray.php
@@ -2,14 +2,20 @@
 
 namespace Laminas\Validator;
 
+use BackedEnum;
 use RecursiveArrayIterator;
 use RecursiveIteratorIterator;
+use UnitEnum;
 
+use function array_map;
 use function in_array;
+use function interface_exists;
+use function is_array;
 use function is_bool;
 use function is_float;
 use function is_int;
 use function is_string;
+use function is_subclass_of;
 
 class InArray extends AbstractValidator
 {
@@ -42,7 +48,7 @@ class InArray extends AbstractValidator
     /**
      * Haystack of possible values
      *
-     * @var array
+     * @var array|class-string<UnitEnum>
      */
     protected $haystack;
 
@@ -66,7 +72,7 @@ class InArray extends AbstractValidator
     /**
      * Returns the haystack option
      *
-     * @return mixed
+     * @return array|class-string<UnitEnum>
      * @throws Exception\RuntimeException If haystack option is not set.
      */
     public function getHaystack()
@@ -80,11 +86,21 @@ class InArray extends AbstractValidator
     /**
      * Sets the haystack option
      *
-     * @param  mixed $haystack
+     * @param array|UnitEnum|BackedEnum $haystack
      * @return $this Provides a fluent interface
      */
-    public function setHaystack(array $haystack)
+    public function setHaystack($haystack)
     {
+        if (! is_array($haystack)) {
+            if (! is_string($haystack) || ! interface_exists(UnitEnum::class)) {
+                throw new Exception\RuntimeException('haystack can only be an array');
+            }
+
+            if (! is_subclass_of($haystack, UnitEnum::class)) {
+                throw new Exception\RuntimeException('haystack has invalid type');
+            }
+        }
+
         $this->haystack = $haystack;
         return $this;
     }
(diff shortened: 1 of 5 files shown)


src/InArray.php
```diff
@@ -42,7 +48,7 @@ class InArray extends AbstractValidator
     /**
      * Haystack of possible values
      *
-     * @var array
+     * @var array|class-string<UnitEnum>
```
Ocramius, line 51: Ok, from a preliminary observation, I think we should leave the haystack alone, and just populate it from a new option containing the enum.

Changing the haystack itself is probably more trouble, especially for BC.

This also means having a new private `setEnum()` or such 🤔

Ocramius (changes requested):
Ok, from a preliminary observation, I think we should leave the haystack alone, and just populate it from a new option containing the enum.

Changing the haystack itself is probably more trouble, especially for BC.

This also means having a new private `setEnum()` or such 🤔

reinfi:
Still Psalm will fail, but I have not really an idea how to handle that because I had nothing todo with psalm so far 🙈

src/InArray.php
```diff
@@ -89,6 +100,30 @@ public function setHaystack(array $haystack)
...
+
+    /**
+     * Set the enum option
+     *
+     * @param class-string<UnitEnum> $enum
+     * @return $this Provides a fluent interface
+     */
+    public function setEnum(string $enum): self
+    {
+        if (! is_subclass_of($enum, UnitEnum::class)) {
```
Ocramius, line 119: I think we can skip this particular check: `class-string<UnitEnum>` already verifies this, and code written in newer codebases generally respects types in a much cleaner way

src/InArray.php
```diff
@@ -46,6 +50,13 @@ class InArray extends AbstractValidator
      */
     protected $haystack;
 
+    /**
+     * Enum of possible values
+     *
+     * @var class-string<UnitEnum>|null
+     */
+    protected $enum;
```
Ocramius, line 58: Let's make this `private` before it spreads :D

src/InArray.php
```diff
@@ -169,6 +204,35 @@ public function setRecursive($recursive)
      */
     public function isValid($value)
     {
+        $enum = $this->getEnum();
```
Ocramius, line 207: Since we make `$this->enum` `private`, we can directly access it here: no need for a method call

src/InArray.php
```diff
@@ -169,6 +204,35 @@ public function setRecursive($recursive)
...
+        if (is_string($enum)) {
```
Ocramius, line 208: We should just check for  `!== null`

src/InArray.php
```diff
@@ -89,6 +100,30 @@ public function setHaystack(array $haystack)
...
+     */
+    public function setEnum(string $enum): self
+    {
+        if (! is_subclass_of($enum, UnitEnum::class)) {
+            throw new Exception\RuntimeException('enum has invalid type');
+        }
+
+        $this->enum = $enum;
+        return $this;
+    }
```
Ocramius, line 125: I wonder if this method should call `setHaystack()`: that would remove the need for touching the `isValid()` implementation

Ocramius (commented):
Overall, approach seems OK 👍 

What I'm thinking about is how to get static analysis "happy" with this :D 

 > Still Psalm will fail, but I have not really an idea how to handle that because I had nothing todo with psalm so far 🙈

We could use a `.stub` file with Psalm, for now, perhaps: we'd declare `BackedEnum` there, then drop the stub once we drop PHP 8.0 support 🤔 

/cc @gsteel do you have any better ideas?

boesing:
@Ocramius one things we need to keep in mind is that filters and stuff are usually passed as attributes and/or annotations. I am working with these Interfaces for years now and I have never called setters directly. They are always called from provided factories from within this component and thus I only have to use the attributes/annotations which allow devs to declare a spec which then is processed. Thats why the abstract filter is used because options are being passed as array so that they can be passed to the plugin manager. thats why the "build" method has that optional options array.

imho, as long as we do not provide an alternative to that specification, we should not try to avoid the Standard which was declared years back. 
So we should not rely on pseudo-types like class-string too much in validator/filter too much. 

----

so I think what would be fine is that we provide a dedicated factory for the ToEnum filter @reinfi, yes. That should work - might be worth testing that in a real project before merging this tho. 😬

PR #169, opened 2023-01-12
Fix #135: better GPS point validator detection of invalid coordinate numeric conversions 

### LINKED ISSUE #135: GpsPoint validator needs a rewrite because it returns true for the input "foo, bar"
> I believe this validator should either be rewritten or removed entirely as it will return true for the input "foo, bar"
> 
> It expects a string as input in the form "latitude, longitude" :
> 
> ```
>     public function isValid($value)
>     {
>         if (strpos($value, ',') === false) {
>             $this->error(self::INCOMPLETE_COORDINATE, $value);
>             return false;
>         }
> 
>         [$lat, $long] = explode(',', $value);
> ```
> 
>  but the `isValidCoordinate()` method blindly casts the supposed lat or long to a double:
> 
> ```
>         $doubleLatitude = (double) $value;
> 
>         if ($doubleLatitude <= $maxBoundary && $doubleLatitude >= $maxBoundary * -1) {
>             return true;
>         }
> ```
> 
> which results in a 0 value in the case of non-numerical strings, which then passes validation as a valid lat/long.
> 
> At the very least it should have explicit comments stating that the expected input is in a very particular format already.

_Originally posted by @ajgale in https://github.com/laminas/laminas-validator/issues/69#issuecomment-1162331488_

codisart:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes
| BC Break      | yes

### Description

This PR is trying to fix the issue #135. 
The fix is to have the string "foo" not converted to 0.0 and validated as a correct coordinate.

### DIFF
--- a/src/GpsPoint.php
+++ b/src/GpsPoint.php
@@ -15,11 +15,10 @@ final class GpsPoint extends AbstractValidator
     public const CONVERT_ERROR         = 'gpsPointConvertError';
     public const INCOMPLETE_COORDINATE = 'gpsPointIncompleteCoordinate';
 
-    /** @var array */
-    protected $messageTemplates = [
-        'gpsPointOutOfBounds'          => '%value% is out of Bounds.',
-        'gpsPointConvertError'         => '%value% can not converted into a Decimal Degree Value.',
-        'gpsPointIncompleteCoordinate' => '%value% did not provided a complete Coordinate',
+    protected array $messageTemplates = [
+        self::OUT_OF_BOUNDS         => '%value% is out of Bounds.',
+        self::CONVERT_ERROR         => '%value% can not converted into a Decimal Degree Value.',
+        self::INCOMPLETE_COORDINATE => '%value% did not provided a complete Coordinate',
     ];
 
     /**
@@ -29,11 +28,9 @@ final class GpsPoint extends AbstractValidator
      * getMessages() will return an array of messages that explain why the
      * validation failed.
      *
-     * @param  mixed $value
-     * @return bool
      * @throws Exception\RuntimeException If validation of $value is impossible.
      */
-    public function isValid($value)
+    public function isValid(mixed $value): bool
     {
         if (! str_contains($value, ',')) {
             $this->error(self::INCOMPLETE_COORDINATE, $value);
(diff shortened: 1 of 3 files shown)


src/GpsPoint.php
```diff
@@ -63,19 +53,17 @@ private function isValidCoordinate($value, float $maxBoundary): bool
             $value = $this->removeDegreeSign($value);
         }
 
-        if ($value === false || $value === null) {
+        if ($value === false || $value === null || !is_numeric($value)) {
```
Ocramius: This ternary can be simplified to just the `is_numeric()` check

test/GPSPointTest.php
```diff
@@ -66,7 +67,7 @@ public function testErrorsSetOnOccur(string $value, string $messageKey, string $
     }
 
     /**
-     * @psalm-return array<array-key, array{0: string}>
+     * @psalm-return array<array-key, arrayG{0: string}>
```
Ocramius: A `G` slipped in

Ocramius (changes requested):
Please check CI failures too

Ocramius:
As this is a bugfix, it should target `2.29.x`

test/GPSPointTest.php
```diff
@@ -85,9 +86,11 @@ public function basicDataProvider(): array
     public function errorMessageTestValues(): array
     {
         return [
-            ['63 47 24.691 N, 18 2 54.363 W', GpsPoint::OUT_OF_BOUNDS, '63 47 24.691 N'],
+            ['63 47 24.691 N, 18 2 54.363 W', GpsPoint::CONVERT_ERROR, '63 47 24.691 N'],
+            ['65° 4\' N,-22.728867530822754', GpsPoint::CONVERT_ERROR, '65° 4\' N'],
```
froschdesign: The previous test was fine, so out of bounds cannot be a conversion error now!

froschdesign: See my comment on the related issue report: https://github.com/laminas/laminas-validator/issues/135#issuecomment-1165290208

test/GPSPointTest.php
codisart: Thank you, I removed it

test/GPSPointTest.php
codisart: Hello @froschdesign, this change in the provider is a mistake for this PR.
I have a new fix where I use the casting to float to determine is the previous string was a partial number.

codisart: Hello @froschdesign, this change in the provider is a mistake for this PR.
I have a new fix where I use the casting to float to determine is the previous string was a partial number before the casting.

src/GpsPoint.php
codisart: After the comment of froschdesign, this check has became obsolete.

Ocramius (approved):
Thanks @codisart!

PR #171, opened 2023-01-23
chore: Convert Date validator to php8 syntax and remove dead code.

codisart:

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

It seems to me that the DateValidator would profit of the php8 syntax. And I also detected some dead code that could be removed. And in the process I also corrected CS errors and warnings.

### DIFF
--- a/src/Date.php
+++ b/src/Date.php
@@ -7,6 +7,7 @@ namespace Laminas\Validator;
 use DateTime;
 use DateTimeImmutable;
 use DateTimeInterface;
+use stdClass;
 use Traversable;
 
 use function array_shift;
@@ -39,22 +40,20 @@ class Date extends AbstractValidator
      *
      * @var string[]
      */
-    protected $messageTemplates = [
+    protected array $messageTemplates = [
         self::INVALID      => 'Invalid type given. String, integer, array or DateTime expected',
         self::INVALID_DATE => 'The input does not appear to be a valid date',
         self::FALSEFORMAT  => "The input does not fit the date format '%format%'",
     ];
 
     /** @var string[] */
-    protected $messageVariables = [
+    protected array $messageVariables = [
         'format' => 'format',
     ];
 
-    /** @var string */
-    protected $format = self::FORMAT_DEFAULT;
+    protected string $format = self::FORMAT_DEFAULT;
 
-    /** @var bool */
-    protected $strict = false;
+    protected bool $strict = false;
 
     /**
      * Sets validator options
@@ -67,6 +66,7 @@ class Date extends AbstractValidator
             $options = iterator_to_array($options);
         } elseif (! is_array($options)) {
             $options        = func_get_args();
+            $temp           = [];
             $temp['format'] = array_shift($options);
             $options        = $temp;
         }
@@ -76,10 +76,8 @@ class Date extends AbstractValidator
 
     /**
      * Returns the format option
-     *
-     * @return string
      */
-    public function getFormat()
+    public function getFormat(): string
     {
         return $this->format;
     }
@@ -114,10 +112,9 @@ class Date extends AbstractValidator
     /**
      * Returns true if $value is a DateTimeInterface instance or can be converted into one.
      *
-     * @param  string|numeric|array|DateTimeInterface $value
-     * @return bool
+     * @param string|int|float|array|stdClass|DateTimeInterface|null $value
      */
-    public function isValid($value)
+    public function isValid($value): bool
     {
         $this->setValue($value);
 
@@ -138,11 +135,10 @@ class Date extends AbstractValidator
     /**
      * Attempts to convert an int, string, or array to a DateTime object
      *
-     * @param string|numeric|array|DateTimeInterface $param
-     * @param bool $addErrors
+     * @param string|int|float|array|stdClass|DateTimeInterface|null $param
      * @return false|DateTime
      */
-    protected function convertToDateTime($param, $addErrors = true)
+    protected function convertToDateTime($param): bool|DateTime
     {
         if ($param instanceof DateTime) {
             return $param;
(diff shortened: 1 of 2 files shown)


src/Date.php
```diff
@@ -39,22 +40,20 @@ class Date extends AbstractValidator
...
         'format' => 'format',
     ];
 
-    /** @var string */
-    protected $format = self::FORMAT_DEFAULT;
+    protected string $format = self::FORMAT_DEFAULT;
 
-    /** @var bool */
-    protected $strict = false;
+    protected bool $strict = false;
```
gsteel, line 56: Adding property type hints to protected properties are a BC break for classes that are open to extension

src/Date.php
```diff
@@ -76,10 +76,8 @@ public function __construct($options = [])
 
     /**
      * Returns the format option
-     *
-     * @return string
      */
-    public function getFormat()
+    public function getFormat(): string
```
gsteel, line 80: BC Break

src/Date.php
```diff
@@ -114,10 +112,9 @@ public function isStrict(): bool
     /**
      * Returns true if $value is a DateTimeInterface instance or can be converted into one.
      *
-     * @param  string|numeric|array|DateTimeInterface $value
-     * @return bool
+     * @param string|int|float|array|stdClass|DateTimeInterface|null $value
      */
-    public function isValid($value)
+    public function isValid($value): bool
```
gsteel, line 117: BC Break

src/Date.php
```diff
@@ -138,11 +135,10 @@ public function isValid($value)
     /**
      * Attempts to convert an int, string, or array to a DateTime object
      *
-     * @param string|numeric|array|DateTimeInterface $param
-     * @param bool $addErrors
+     * @param string|int|float|array|stdClass|DateTimeInterface|null $param
      * @return false|DateTime
      */
-    protected function convertToDateTime($param, $addErrors = true)
+    protected function convertToDateTime($param): bool|DateTime
```
gsteel, line 141: Method signature change is a BC break and the flag to add errors during internal conversion is not dead code, even if it seems to be always `true`, it may not be during inheritance.

src/Date.php
```diff
@@ -152,55 +148,38 @@ protected function convertToDateTime($param, $addErrors = true)
...
 
     /**
      * Attempts to convert a string into a DateTime object
-     *
-     * @param string $value
-     * @param bool $addErrors
-     * @return false|DateTime
      */
-    protected function convertString($value, $addErrors = true)
+    protected function convertString(string $value): false|DateTime
```
gsteel, line 182: BC Break

src/Date.php
```diff
@@ -152,55 +148,38 @@ protected function convertToDateTime($param, $addErrors = true)
...
 
     /**
-     * Attempts to convert an double into a DateTime object
-     *
-     * @param double $value
-     * @return false|DateTime
+     * Attempts to convert an number (float or integer) into a DateTime object
      */
-    protected function convertDouble($value)
+    protected function convertNumber(int|float $value): false|DateTime
```
gsteel, line 174: BC Break

src/Date.php
```diff
@@ -152,55 +148,38 @@ protected function convertToDateTime($param, $addErrors = true)
...
-     * @param integer $value
-     * @return false|DateTime
+     * @return false
      */
-    protected function convertInteger($value)
+    private function addInvalidTypeError(): bool
     {
-        return DateTime::createFromFormat('U', (string) $value);
+        $this->error(self::INVALID);
+        return false;
```
gsteel, line 168: Adding a private method is fine, but the removal of the protected method is a BC break

src/Date.php
```diff
@@ -224,12 +201,10 @@ protected function convertString($value, $addErrors = true)
...
      * Implodes the array into a string and proxies to {@link convertString()}.
      *
-     * @param bool $addErrors
-     * @return false|DateTime
-     * @todo   enhance the implosion
+     * @todo enhance the implosion
      */
-    protected function convertArray(array $value, $addErrors = true)
+    protected function convertArray(array $value): false|DateTime
     {
```
gsteel, line 207: BC break

src/Date.php
```diff
@@ -152,55 +148,38 @@ protected function convertToDateTime($param, $addErrors = true)
...
-            $this->error(self::INVALID);
-        }
-
-        return false;
+        /**
+         * @psalm-suppress PossiblyInvalidArgument
+         */
+        return match (gettype($param)) {
+            'string' => $this->convertString($param),
+            'integer', 'float' => $this->convertNumber($param),
```
gsteel, line 156: `gettype` on a float will yield `double`: https://3v4l.org/XG0Cj

src/Date.php
```diff
@@ -152,55 +148,38 @@ protected function convertToDateTime($param, $addErrors = true)
...
-        }
-
-        if ($addErrors) {
-            $this->error(self::INVALID);
-        }
-
-        return false;
+        /**
+         * @psalm-suppress PossiblyInvalidArgument
+         */
```
gsteel, line 153: This suppression shouldn't be necessary

src/DateStep.php
```diff
@@ -48,7 +48,7 @@ class DateStep extends Date
      *
      * @var string[]
      */
-    protected $messageTemplates = [
+    protected array $messageTemplates = [
```
gsteel, line 51: BC break

src/DateStep.php
```diff
@@ -182,12 +182,8 @@ public function setTimezone(DateTimeZone $timezone)
      * Supports formats with ISO week (W) definitions
      *
      * @see Date::convertString()
-     *
-     * @param string $value
-     * @param bool $addErrors
-     * @return DateTime|false
      */
-    protected function convertString($value, $addErrors = true)
+    protected function convertString(string $value): DateTime|false
```
gsteel, line 186: BC break

src/DateStep.php
```diff
@@ -204,9 +200,6 @@ protected function convertString($value, $addErrors = true)
         // and still return a DateTime object.
         $errors = DateTime::getLastErrors();
         if (is_array($errors) && $errors['warning_count'] > 0) {
-            if ($addErrors) {
-                $this->error(self::FALSEFORMAT);
-            }
```
gsteel, line 209: BC break

src/DateStep.php
```diff
@@ -217,17 +210,16 @@ protected function convertString($value, $addErrors = true)
      * Returns true if a date is within a valid step
      *
      * @param string|int|DateTimeInterface $value
-     * @return bool
      * @throws Exception\InvalidArgumentException
      */
-    public function isValid($value)
+    public function isValid($value): bool
```
gsteel, line 215: BC break

gsteel (changes requested):
Unfortunately, due to possible inheritance, the vast majority of this patch breaks BC. It's only possible to make these changes by targeting a major release, and would therefore also need documentation of the breaking changes.

Ocramius:
Will close here: while the intent of the patch is admirable, it is more work to adjust it, than to re-open a new patch with more minimal changes.

src/Date.php
codisart: Oh so maybe we are missing a test ?

PR #172, opened 2023-01-26
Introduce `IsJsonString` validator

gsteel:
|    Q          |   A
|-------------- | ------
| New Feature   | yes

### Description

Adds a straight-forward validator to help determine if a string would successfully decode to JSON.

Provides options for limiting acceptable input, or rather the acceptable types the json would decode to and an option for max depth.

I was surprised to not find something like this already here!

### DIFF
--- a/docs/book/validators/is-json-string.md
+++ b/docs/book/validators/is-json-string.md
@@ -1,4 +1,4 @@
-# IsJsonString Validator
+# IsJsonString
 
 `Laminas\Validator\IsJsonString` allows you to validate whether a given value is a string that will be successfully decoded by `json_decode`.
 
@@ -15,7 +15,7 @@ if ($validator->isValid($input)) {
 }
 ```
 
-## Restricting Acceptable JSON types
+## Restricting Acceptable JSON Types
 
 `json_decode` accepts numeric strings representing integers and floating point numbers, booleans, arrays and objects.
 You can restrict what is considered valid input using the `allow` option of the validator.
@@ -32,9 +32,9 @@ $validator->isValid('true'); // false
 
 The `allow` option is a bit mask of the `ALLOW_*` constants in `IsJsonString`:
 
-- `IsJsonString::ALLOW_INT` - Accept numeric such as "1"
-- `IsJsonString::ALLOW_FLOAT` - Accept numeric strings such as "1.234"
-- `IsJsonString::ALLOW_BOOL` - Accept "true" and "false"
+- `IsJsonString::ALLOW_INT` - Accept integer such as `1`
+- `IsJsonString::ALLOW_FLOAT` - Accept floating-point value such as `1.234`
+- `IsJsonString::ALLOW_BOOL` - Accept `true` and `false`
 - `IsJsonString::ALLOW_ARRAY` - Accept JSON arrays such as `["One", "Two"]`
 - `IsJsonString::ALLOW_OBJECT` - Accept JSON objects such as `{"Some":"Object"}`
 - `IsJsonString::ALLOW_ALL` - A convenience constant allowing all of the above _(Also the default)_.
@@ -46,7 +46,7 @@ use Laminas\Validator\IsJsonString;
 
 $validator = new IsJsonString();
 $validator->setAllow(IsJsonString::ALLOW_ARRAY | IsJsonString::ALLOW_OBJECT);
-$validator->isValid("1.234"); // false
+$validator->isValid('1.234'); // false
 ```
 
 ## Restricting Max Object or Array Nesting Level
--- a/docs/book/validators/is-json-string.md
+++ b/docs/book/validators/is-json-string.md
@@ -1,0 +1,66 @@
+# IsJsonString Validator
+
+`Laminas\Validator\IsJsonString` allows you to validate whether a given value is a string that will be successfully decoded by `json_decode`.
+
+## Basic Usage
+
+```php
+$validator = new Laminas\Validator\IsJsonString();
+$input = '{"some":"json"}';
+
+if ($validator->isValid($input)) {
+    // $input can be successfully decoded
+} else {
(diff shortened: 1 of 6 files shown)


boesing:
Can we probably keep maintenance in dedicated tickets rather than adding them to a feature?
Since we do not put effort into changelogs and thus these usually contain the PR title, as a user, I'd not expect anything else to change but having a new validator available.

This PR does plenty of other stuff (such as refining types, etc.) and thus I'd really encourage you to put that QA stuff into a dedicated PR.

gsteel:
Acknowledged @boesing - I was just trying to get CI green here on unrelated issues. I'll revert to the bare minimum to get a thumbs up from psalm on locked deps tomorrow.

froschdesign:
> Since we do not put effort into changelogs and thus these usually contain the PR title…

**This is absolutely correct!** 

If we look at [the latest release notes of laminas-view](https://github.com/laminas/laminas-view/releases/tag/2.26.0), the information they contain is poor and useless for a user:

> ### Enhancement
> Remove laminas/console dev dependency and stop ignoring platform requirements

Does this mean that something no longer works and why was this removed in a minor version?

> ### Documentation,Enhancement
> Merge release 2.25.1 into 2.26.x
> ### Bug,Documentation
> Merge release 2.25.2 into 2.26.x

What does that mean? Does this bring me anything as a user or why is this information included in a release note?

---

My point is that meaningful information must be included and also nothing must be left out, like the other changes in this pull request.

froschdesign:
@gsteel 
I like it a lot and also that different types can be allowed! 👍🏻

docs/book/validators/is-json-string.md
```diff
@@ -0,0 +1,66 @@
+# IsJsonString Validator
```
froschdesign: The first headline must be synchronised with the navigation entry.

```suggestion
# IsJsonString
```

_(Right, that's not true for most of the other pages.)_

docs/book/validators/is-json-string.md
```diff
@@ -0,0 +1,66 @@
...
+$input = '{"some":"json"}';
+
+if ($validator->isValid($input)) {
+    // $input can be successfully decoded
+} else {
+    // $input is not a valid JSON string
+}
+```
+
+## Restricting Acceptable JSON types
```
froschdesign:
```suggestion
## Restricting Acceptable JSON Types
```

docs/book/validators/is-json-string.md
```diff
@@ -0,0 +1,66 @@
...
+]);
+
+$validator->isValid('true'); // false
+```
+
+The `allow` option is a bit mask of the `ALLOW_*` constants in `IsJsonString`:
+
+- `IsJsonString::ALLOW_INT` - Accept numeric such as "1"
+- `IsJsonString::ALLOW_FLOAT` - Accept numeric strings such as "1.234"
+- `IsJsonString::ALLOW_BOOL` - Accept "true" and "false"
```
froschdesign:
```suggestion
- `IsJsonString::ALLOW_INT` - Accept integer such as `1`
- `IsJsonString::ALLOW_FLOAT` - Accept floating-point value such as `1.234`
- `IsJsonString::ALLOW_BOOL` - Accept `true` and `false`
```

docs/book/validators/is-json-string.md
```diff
@@ -0,0 +1,66 @@
...
+- `IsJsonString::ALLOW_ALL` - A convenience constant allowing all of the above _(Also the default)_.
+
+The `allow` option also has a companion setter method `setAllow`. For example, to only accept arrays and objects:
+
+```php
+use Laminas\Validator\IsJsonString;
+
+$validator = new IsJsonString();
+$validator->setAllow(IsJsonString::ALLOW_ARRAY | IsJsonString::ALLOW_OBJECT);
+$validator->isValid("1.234"); // false
```
froschdesign:
```suggestion
$validator->isValid('1.234'); // false
```

froschdesign (changes requested):

froschdesign (approved):
Looks good to me! 👍🏻 

(I have already updated the milestone description.)

src/IsJsonString.php
```diff
@@ -0,0 +1,152 @@
...
+ * @psalm-type Options = AbstractOptions|CustomOptions
+ */
+final class IsJsonString extends AbstractValidator
+{
+    public const ERROR_NOT_STRING         = 'errorNotString';
+    public const ERROR_TYPE_NOT_ALLOWED   = 'errorTypeNotAllowed';
+    public const ERROR_MAX_DEPTH_EXCEEDED = 'errorMaxDepthExceeded';
+    public const ERROR_INVALID_JSON       = 'errorInvalidJson';
+
+    public const ALLOW_INT    = 0b0000001;
```
Ocramius, line 35: IMO the validator is called "IsJsonString", so if non-string input is given, it should just fail, no other complexity needed (ambiguous otherwise)

Ocramius (changes requested):

src/IsJsonString.php
gsteel: It does fail when given a non string - the 'allow' represents JSON strings in the form "1.2", "1", "true", "false", "[]", and "{}"

Ocramius: Hmm, I'm not convinced we should do this sort of validation in here 🤔

Anyway, if "allow all" is the default, I'm OK with this patch (sorry, only doing reviews from my phone, while on the run 😬)

gsteel: Allow all is the default. The general idea here is "Will it decode or not?" - The ALLOW_* constants allow you fine tune the guard so that you can say, "I only want JSON objects to pass validation" fe.

Ocramius (approved):
@gsteel if you were waiting for me to merge, sorry, here's my approval meanwhile :D

Issue #85, opened 2021-01-13
[RFC]: Hybrid options transform system


codisart:
## RFC

| Q                   | A      |
| ------------------- | ------ |
| Proposed Version(s) | 2.15.x  |
| BC Break?           | No |

### Goal

Remove duplication and simplify for the initialisations of the option system.

### Background

Several validator objects use an hybrid system of options. 
It can be passed as an single array argument or as multiples arguments of the constructor
So a part of the job of the constructor is to transform options into an array. 
And several validators do this job and it is not easy to understand and to read.

### Considerations

I expect a good impact on maintainers due to this change. 
There should be no impact on current behavior.

### Proposal(s)

The proposal is to use a class with a static factory method to init the options array 
This method will accept:
as first argument an array of the options keys  that always should be string
as second argument an array that should always be the array returned by the function func_get_args()

```php
$options = ValidatorOptions::asArray(['min', 'max', 'inclusive'], func_get_args());
```
```php
class ValidatorOptions
{
    public static asArray(array $keys, array $args) {
        $this->shouldHaveOnlyStringValues($keys);

        $result = [];
        foreach ($args as $arg) {
            $result[array_shift($keys)] = $arg;
        }
        return $result;
    }

    private function shouldHaveOnlyStringValues(array $keys) : void
    {
        foreach ($keys as $key) {
            if (! is_string($key)) {
                throw new \InvalidArgumentException('The values of $keys should be only strings');
            }
        }
    }
}
```
### Appendix 

I'd like to thank @glensc for the final form of the proposal

Here the PR with the two first forms of the proposed solution
https://github.com/laminas/laminas-validator/pull/60

gsteel:
Interesting idea, but I personally would prefer to see only array shapes for options, obviously as a BC breaking change for the next major.

A lot of the validators accept an options array or a specific option as argument 1, such as `Between` which is effecively `Between::__construct(array|int $minOrOptions, int|null $max, bool|null $inclusive)`. This is difficult to document and maintain and I'd much rather see this:

```php

/**
 * @psalm-type Options = array{
 *    option1: non-empty-string,
 *    option2: int<1, 42>,
 *    optional?: bool,
 * }
 */
final class SomeValidator implements ValidatorInterface
{
    /** @param Options $options */
    public function __construct(private array $options)
    {
    }
}
```

I'd also be in favour of removing `getOptions`, `setOptions`, all setters and getters for individual options, and the idea that options arrays can be `Traversables` too.

[...]

PR #177, opened 2023-02-19
chore: Convert some parts of Date validator to php8 syntax and add some tests to check userland retrocompatibility.

codisart:

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

It seems to me that the DateValidator would profit of the php8 syntax. 
I have also added some tests to allow a check on userland breaking changes.

### DIFF
--- a/src/Date.php
+++ b/src/Date.php
@@ -67,6 +67,7 @@ class Date extends AbstractValidator
             $options = iterator_to_array($options);
         } elseif (! is_array($options)) {
             $options        = func_get_args();
+            $temp           = [];
             $temp['format'] = array_shift($options);
             $options        = $temp;
         }
@@ -152,22 +153,23 @@ class Date extends AbstractValidator
             return DateTime::createFromImmutable($param);
         }
 
-        $type = gettype($param);
-        switch ($type) {
-            case 'string':
-                return $this->convertString($param, $addErrors);
-            case 'integer':
-                return $this->convertInteger($param);
-            case 'double':
-                return $this->convertDouble($param);
-            case 'array':
-                return $this->convertArray($param, $addErrors);
-        }
+        return match (gettype($param)) {
+            'string' => $this->convertString($param, $addErrors),
+            'integer' => $this->convertInteger($param),
+            'double' => $this->convertDouble($param),
+            'array' => $this->convertArray($param),
+            default => $this->addInvalidTypeError($addErrors),
+        };
+    }
 
+    /**
+     * @return false
+     */
+    private function addInvalidTypeError(bool $addErrors)
+    {
         if ($addErrors) {
             $this->error(self::INVALID);
         }
-
         return false;
     }
 
--- a/test/TestAsset/CustomDate.php
+++ b/test/TestAsset/CustomDate.php
@@ -1,0 +1,55 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LaminasTest\Validator\TestAsset;
+
+use DateTime;
+use DateTimeInterface;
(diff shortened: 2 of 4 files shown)


src/Date.php
```diff
@@ -152,22 +153,23 @@ protected function convertToDateTime($param, $addErrors = true)
...
+            'string' => $this->convertString($param, $addErrors),
+            'integer' => $this->convertInteger($param),
+            'double' => $this->convertDouble($param),
+            'array' => $this->convertArray($param),
+            default => $this->addInvalidTypeError($addErrors),
+        };
+    }
 
+    /**
+     * @return false
```
Ocramius, line 166: Could be a native type declaration

src/Date.php
```diff
@@ -67,6 +67,7 @@ public function __construct($options = [])
             $options = iterator_to_array($options);
         } elseif (! is_array($options)) {
             $options        = func_get_args();
+            $temp           = [];
             $temp['format'] = array_shift($options);
```
Ocramius, line 71: Since this is probably part of the most common happy-path of most apps, I endorse `$temp = ['format' => array_shift($options)]`, but not that important

test/TestAsset/CustomDate.php
```diff
@@ -0,0 +1,55 @@
...
+    {
+        return parent::isValid($value);
+    }
+
+    /**
+     * @param string|numeric|array|DateTimeInterface $param
+     * @param bool $addErrors
+     * @return bool|DateTime
+     */
+    protected function convertToDateTime($param, $addErrors = true)
```
Ocramius, line 33: Are these just here for BC issues? :thinking:

test/TestAsset/CustomDate.php
```diff
@@ -0,0 +1,55 @@
...
+
+declare(strict_types=1);
+
+namespace LaminasTest\Validator\TestAsset;
+
+use DateTime;
+use DateTimeInterface;
+use Laminas\Validator\Date;
+
+final class CustomDate extends Date
```
Ocramius, line 11: Let's document why we introduced this specific asset

Ocramius (commented):
Looking good:

test/TestAsset/CustomDate.php
codisart: This was the idea, yes.

src/Date.php
codisart: I wanted to keep the same "behaviour"/"pattern" than others options arguments to be easier to find out which are using this. Like Between.php or DateStep.php.

It is actually something that I thought could be changed and I did a RFC for the options.
https://github.com/laminas/laminas-validator/issues/85

src/Date.php
codisart: Is it not only on 8.2 where we can have false as a return type ?

Ocramius: Ah yes! Sorry, indeed

src/Date.php
Ocramius: I'm fine with leaving this as-is, for now :)

gsteel:
@codisart Could you take a look at #330 which is a refactor of the date validator already merged into `3.0.x` and send any improvements you think are necessary in a separate patch and/or close this here? 3.0 should be released relatively soon and whilst this should have been merged ages ago, you might not think this patch is necessary anymore.

gsteel:
Thanks for your work here @codisart but I'm closing this patch as outdated now

PR #179, opened 2023-02-23
Fix value annotation in isValid method for Regex validator.

h3nnry:
Changed annotation for parameter value in isValid method for Regex validator, from string to string|numerical as long as this method accepts also integer and float parameters.

### DIFF
--- a/src/Regex.php
+++ b/src/Regex.php
@@ -107,7 +107,7 @@ class Regex extends AbstractValidator
     /**
      * Returns true if and only if $value matches against the pattern option
      *
-     * @param  string $value
+     * @param  string|numeric $value
      * @return bool
      */
     public function isValid($value)
@@ -120,6 +120,7 @@ class Regex extends AbstractValidator
         $this->setValue($value);
 
         ErrorHandler::start();
+        /** @psalm-suppress PossiblyInvalidArgument */
         $status = preg_match($this->pattern, $value);
         ErrorHandler::stop();
         if (false === $status) {
--- a/src/Regex.php
+++ b/src/Regex.php
@@ -117,10 +117,13 @@ class Regex extends AbstractValidator
             return false;
         }
 
+        if (is_int($value) || is_float($value)) {
+            $value = (string) $value;
+        }
+
         $this->setValue($value);
 
         ErrorHandler::start();
-        /** @psalm-suppress PossiblyInvalidArgument */
         $status = preg_match($this->pattern, $value);
         ErrorHandler::stop();
         if (false === $status) {
--- a/test/RegexTest.php
+++ b/test/RegexTest.php
@@ -213,24 +213,27 @@ final class RegexTest extends TestCase
     /**
      * @dataProvider numericDataProvider
      */
-    public function testNumbers(int|float $input, bool $expected): void
+    public function testNumbers(string $pattern, int|float $input, bool $expected): void
     {
-        $validator = new Regex('/^(-?\d+(?:\.\d+)?+)$/');
+        $validator = new Regex($pattern);
 
         self::assertSame($expected, $validator->isValid($input));
     }
 
     /**
-     * @psalm-return array<array-key, array{0: int|float, 1: bool}>
+     * @psalm-return array<array-key, array{0: string, 1: int|float, 2: bool}>
      */
-    public function numericDataProvider(): array
+    public static function numericDataProvider(): array
     {
         return [
-            [12345, true],
-            [PHP_INT_MAX, true],
-            [-123, true],
-            [0.0099, true],
-            [-100.50, true],
+            ['/^(-?\d+(?:\.\d+)?+)$/', 12345, true],
+            ['/^(-?\d+(?:\.\d+)?+)$/', PHP_INT_MAX, true],
+            ['/^(-?\d+(?:\.\d+)?+)$/', -123, true],
+            ['/^(-?\d+(?:\.\d+)?+)$/', 0.0099, true],
+            ['/^(-?\d+(?:\.\d+)?+)$/', -100.50, true],
+            ['/^\d+$/', 100, true],
+            ['/^\d+$/', -100, false],
+            ['/^\d+$/', 123.45, false],
         ];
     }
 }
--- a/test/RegexTest.php
+++ b/test/RegexTest.php
@@ -12,6 +12,8 @@ use ReflectionProperty;
 use function array_keys;
 use function implode;
 
+use const PHP_INT_MAX;
+
 /**
  * @group Laminas_Validator
  * @covers \Laminas\Validator\Regex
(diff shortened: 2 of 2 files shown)


src/Regex.php
```diff
@@ -120,6 +120,7 @@ public function isValid($value)
         $this->setValue($value);
 
         ErrorHandler::start();
+        /** @psalm-suppress PossiblyInvalidArgument */
```
Ocramius: Not acceptable here

src/Regex.php
```diff
@@ -107,7 +107,7 @@ public function setPattern($pattern)
     /**
      * Returns true if and only if $value matches against the pattern option
      *
-     * @param  string $value
+     * @param  string|numeric $value
```
Ocramius, line 110: If we accept non-string values, a string cast is probably needed for acceptable conversions

Ocramius (changes requested):

src/Regex.php
h3nnry: removed.

src/Regex.php
h3nnry: added.

src/Regex.php
```diff
@@ -117,6 +117,10 @@ public function isValid($value)
             return false;
         }
 
+        if (is_int($value) || is_float($value)) {
+            $value = (string) $value;
+        }
```
Ocramius, line 122: Since new logic was added, we should have a test covering this scenario?

Ocramius (changes requested):

src/Regex.php
h3nnry: added

Ocramius (approved):
LGTM, although asking for another review by others.

src/Regex.php
internalsystemerror: I have reservations against allowing floats for the same reasons it was removed from moneyphp (I believe by @Ocramius), and as was discussed here: https://github.com/moneyphp/money/issues/650.

It would then be the responsibility of the user to handle that casting.
```php
$regExValidator->isValid((string)$value);
```

Maybe there needs to be a `ToString` filter for use with [`laminas/laminas-inputfilter`](https://github.com/laminas/laminas-inputfilter)?

Although not as immediately important as with a package named `money`, since this is for validation I could see some users unknowingly making a mistake. Thoughts? Is this an overreaction?

internalsystemerror (commented):
https://github.com/laminas/laminas-validator/pull/179#discussion_r1123668853 is my only reservation. But if people think that's overcautious then I'm happy to approve.

src/Regex.php
Ocramius: Ooooh, that is a good point! A floating point number may indeed cast to something unexpected 🤔

froschdesign: @internalsystemerror 
> Maybe there needs to be a ToString filter…

See: https://github.com/laminas/laminas-filter/issues/98

src/Regex.php
gsteel: IMO, `isValid()` should always accept `mixed` as per [the interface](https://github.com/laminas/laminas-validator/blob/2.31.x/src/ValidatorInterface.php).

src/Regex.php
gsteel: I don't think we can restrict the input type of the validator beyond `mixed` to begin with, so I think that the string cast is necessary to preserve existing behaviour. In hind-sight, the first conditional should have failed on any non-string value, but as it doesn't, at least the cast prevents `int|float` reaching `preg_match`

test/RegexTest.php
```diff
@@ -207,4 +209,28 @@ public function testIsValidShouldReturnFalseWhenRegexPatternIsInvalid(): void
...
+    public function numericDataProvider(): array
+    {
+        return [
+            [12345, true],
+            [PHP_INT_MAX, true],
+            [-123, true],
+            [0.0099, true],
+            [-100.50, true],
+        ];
+    }
```
gsteel: Can we also add some failures here too by adding the pattern to the data provider such as `[123.45, '/^\d+$/', false]`? _(also make it static for PHPUnit 10)_

gsteel (changes requested):

test/RegexTest.php
h3nnry: added.

gsteel: Thanks @h3nnry

src/Regex.php
Ocramius: The only scenario I'm worried about here is a string cast for stuff that ends up being `1.07E-5` or such :thinking:

src/Regex.php
h3nnry: In my opinion better to avoid mixed type.

src/Regex.php
h3nnry: In my opinion it's user who need to avoid scientific notation, and if he's ok with it it need to created the correct pattern.

Ocramius: A `(string)` cast of a `float` value may lead to the scientific notation regardless.

gsteel: Yeah - I get that concern, but there's already an implicit cast to string during `preg_match` anyhow. I think the only way to "fix" this is to reject int|float up-front and I don't think we can get away with that due to existing behaviour.

Perhaps a note should be added to remove the int|float checks for the next major and fail non-string input, making it the callers responsibility to cast if necessary, leaving the current implementation broken in the same predictable way.

My understanding is that this patch does nothing to change the existing behaviour, i.e:

```php
preg_match('/^\d{1}\.\d{2}$/', 0.01); // 1
preg_match('/^\d{1}\.\d{2}$/', '0.01'); // 1
preg_match('/^\d{1}\.\d{2}$/', 35 + -34.99); // 0
preg_match('/^\d{1}\.\d{2}$/', '35' + '-34.99'); // 0
```

internalsystemerror: If it's currently broken because there is no parameter type hint then I don't think we can "fix" it. But we CAN prevent people from making the mistake if they're intending to use static analysis.

I was thinking we use `@param string|int` instead of `@param string|int|float` more than any specific handling using `is_float`. Though now that consider I @gsteel's position, I guess validators will always be required to accept a value of mixed type, so should it be throwing an error (by setting a parameter type hint in a future major), or should it just be returning false?

If the latter then it should stay as mixed and then `if(!is_string($value) && !is_int($value)) { return false; }`?

Ocramius: Beware that the interface does indeed say `mixed`:

https://github.com/laminas/laminas-validator/blob/b904e5d197633e0eeb4c0162d8e7604e8829d10a/src/ValidatorInterface.php#L22-L26

gsteel:
Thanks for your work here @h3nnry - #280 has been merged into 3.0.x, so I think this PR is no longer relevant.

PR #182, opened 2023-05-18
Remove support for PHP 8.0, Upgrade PHPUnit to 10.1.x

gsteel:
Closes #175

- Bumps Service Manager to ^3.21 for plugin manager tests that are compatible with PHPUnit 10
- Fixes quite a few psalm issues that are eclipsed by the UnusedMethod baseline additions
- Enables MX checks that hit the network by default

### DIFF
--- a/src/StaticValidator.php
+++ b/src/StaticValidator.php
@@ -62,9 +62,7 @@ class StaticValidator
             );
         }
 
-        $plugins = static::getPluginManager();
-
-        /** @var ValidatorInterface $validator */
+        $plugins   = static::getPluginManager();
         $validator = $plugins->get($classBaseName, $options);
 
         return $validator->isValid($value);
--- a/src/StaticValidator.php
+++ b/src/StaticValidator.php
@@ -64,7 +64,9 @@ class StaticValidator
 
         $plugins = static::getPluginManager();
 
+        /** @var ValidatorInterface $validator */
         $validator = $plugins->get($classBaseName, $options);
+
         return $validator->isValid($value);
     }
 }
--- a/src/ValidatorPluginManager.php
+++ b/src/ValidatorPluginManager.php
@@ -4,7 +4,6 @@ namespace Laminas\Validator;
 
 use Laminas\I18n\Validator as I18nValidator;
 use Laminas\ServiceManager\AbstractPluginManager;
-use Laminas\ServiceManager\ConfigInterface;
 use Laminas\ServiceManager\Exception\InvalidServiceException;
 use Laminas\ServiceManager\Factory\InvokableFactory;
 use Laminas\ServiceManager\ServiceManager;
@@ -47,12 +46,8 @@ use function method_exists;
 use function sprintf;
 
 /**
- * @link ConfigInterface
- * @link ServiceManager
- *
  * @psalm-import-type ServiceManagerConfiguration from ServiceManager
- * @template InstanceType of ValidatorInterface
- * @extends AbstractPluginManager<InstanceType>
+ * @extends AbstractPluginManager<ValidatorInterface>
  */
 class ValidatorPluginManager extends AbstractPluginManager
 {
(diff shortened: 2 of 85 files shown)


src/StaticValidator.php
```diff
@@ -64,7 +64,9 @@ public static function execute(mixed $value, $classBaseName, array $options = []
 
         $plugins = static::getPluginManager();
 
+        /** @var ValidatorInterface $validator */
```
Ocramius: This shouldn't be needed

gsteel: For some reason Psalm doesn't like the non-compound `@psalm-type InstanceType of ValidatorInterface`. Changing the plugin manager to `@extends AbstractPluginManager<ValidatorInterface>` seems to clear it up. 🤷‍♂️

gsteel: edit: I just reviewed what was there before and it had `@psalm-template InstanceType of …`, so no wonder type inference wasn't working

src/ValidatorPluginManager.php
```diff
@@ -545,7 +506,7 @@ class ValidatorPluginManager extends AbstractPluginManager
     /**
      * Default instance type
      *
-     * @var string
+     * @inheritDoc
```
internalsystemerror, line 509: There is no parent for this property in `3.21.0`, that would be in `AbstractSingleInstancePluginManager ` not `AbstractPluginManager`.

You could say `@var class-string<ValidatorInterface>`.

src/ValidatorPluginManager.php
```diff
@@ -568,20 +529,16 @@ public function __construct($configOrContainerInstance = null, array $v3config =
     }
 
     /**
-     * Validate plugin instance
-     *
-     * {@inheritDoc}
-     *
      * @param mixed $instance
-     * @psalm-assert InstanceType $instance
+     * @psalm-assert ValidatorInterface $instance
```
internalsystemerror, line 533: I'm not getting why the template usage was removed? Is it not helpful/useful at all?

internalsystemerror (commented):
Tests all look good. It mainly seems to be making static where necessary and using attributes over annotations.

The psalm baseline looks like a lot of changes, however this also appears to be in line with updating the format psalm is using.

A couple of questions regarding changes to src, however. Are we extending the correct abstract class for ValidatorPluginManager, and do we need to lose the use of the generics template?

src/ValidatorPluginManager.php
gsteel: Yes there is - and it's annotated as `class-string<InstanceType>|null` which causes `NonInvariantDocBlockType` issues if you try and tighten it up. `@inheritDoc` works because Psalm figures out it's already non-null value and the template defines the class-string type

src/ValidatorPluginManager.php
gsteel: It means that internally, anywhere you _use_ the plugin manager, psalm can't infer the return type of `get()` unless you annotate each instance as `ValidatorPluginManager<ValidatorInterface>` - I'm not 100% sure on this, that's why I added the SA test. The template is only useful if you're extending from the plugin manager here, at which point, you'd have to override everything so you might as well inherit from the abstract. Really, this plugin manager should be final (I reckon)

src/ValidatorPluginManager.php
internalsystemerror: Oh my bad. Its in the tagged version https://github.com/laminas/laminas-servicemanager/blob/3.21.0/src/AbstractPluginManager.php, but not on the branch https://github.com/laminas/laminas-servicemanager/blob/3.21.x/src/AbstractPluginManager.php.

src/ValidatorPluginManager.php
internalsystemerror: Makes sense. It seems we can't set a template default either, nor will we be able to https://github.com/vimeo/psalm/issues/3048.

internalsystemerror (approved):
LGTM!

src/ValidatorPluginManager.php
gsteel: Thanks for that link - the solution to the problem I was having makes sense now!!

PR #184, opened 2023-05-25
Update IANA TLDs

gsteel:


Ocramius:
Wait, didn't `.zip` and similar also make it in here?

Ocramius:
@gsteel is it possible to somehow automate this process, or at least make CI fail when a diff is generated here? :thinking:

Xerkus:
Shouldn't be hard to automate. Update script is already there so only scheduled workflow is needed to run script and then create or update PR.

gsteel:
I was just thinking that. Added an issue: https://github.com/laminas/laminas-validator/issues/185

gsteel:
This is redundant now #186 is underway

PR #217, opened 2023-12-06
Fix constructor expectations

gsteel:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes

### Description

The constructor should allow an adapter instance and throw an exception for other invalid types.


gsteel:
1 psalm issue added to the baseline - `!is_array` is impossible according to documented types

Ocramius (approved):

PR #129, opened 2022-06-13
QA: Bumps Dependencies and Improves Static Analysis

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

- Drops PHP 7.3
- Drops Service Manager v2
- Updates latest dev dependencies
- Adds template types for plugin manager
- Fixes newly surfaced type errors
- Drops Interop\Container
- Moves Service Manager from dev deps to required deps

PR #232, opened 2024-03-04
QA(fix): Move laminas/service-manager back to dev dependencies

michalbundyra:
Pretty unsure why it was moved from dev dependencies in #129, but it seems to be wrong.

Still suggest section states:

> "laminas/laminas-servicemanager": "Laminas\\ServiceManager component to allow using the ValidatorPluginManager and validator chains",

so it is optional dependency and we should not be forced to install it, when we do not need it.

|    Q          |   A
|-------------- | ------
| Documentation | yes/no
| Bugfix        | yes/no
| BC Break      | yes/no
| New Feature   | yes/no
| RFC           | yes/no
| QA            | yes/no

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,6 @@
     },
     "require": {
         "php": "~8.1.0 || ~8.2.0 || ~8.3.0",
-        "laminas/laminas-servicemanager": "^3.21.0",
         "laminas/laminas-stdlib": "^3.13",
         "psr/http-message": "^1.0.1 || ^2.0.0"
     },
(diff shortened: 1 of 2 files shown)


Xerkus:
ValidatorPluginManager needs it. We should move away from optional dependencies. Especially now that incompatible service manager release was done

gsteel:
@michalbundyra I agree with @Xerkus here. ServiceManager is an important dependency here and we need to restrict the installable version to those that are compatible.

PR #236, opened 2024-03-13
Add `UploadedFileInterface` to valid types of `$value` parameter of `MimeType::isValid()`

akrabat:
The `$value` parameter of `MimeType::isValid()` is passed to [`FileInformationTrait::getFileInfo()`](https://github.com/laminas/laminas-validator/blob/2.51.x/src/File/FileInformationTrait.php#L17) which takes `string|array|object`, so `isValid()`'s `$value` should accept the same types.

### DIFF
--- a/src/File/ExcludeMimeType.php
+++ b/src/File/ExcludeMimeType.php
@@ -36,8 +36,8 @@ class ExcludeMimeType extends MimeType
      * of mimetypes can be checked. If you give for example "image" all image
      * mime types will not be accepted like "image/gif", "image/jpeg" and so on.
      *
-     * @param  string|array $value Real file to check for mimetype
-     * @param  array        $file  File data from \Laminas\File\Transfer\Transfer (optional)
+     * @param  string|array|object $value Real file to check for mimetype
+     * @param  array               $file  File data from \Laminas\File\Transfer\Transfer (optional)
      * @return bool
      */
     public function isValid($value, $file = null)
--- a/src/File/ExcludeMimeType.php
+++ b/src/File/ExcludeMimeType.php
@@ -2,6 +2,8 @@
 
 namespace Laminas\Validator\File;
 
+use Psr\Http\Message\UploadedFileInterface;
+
 use function array_merge;
 use function class_exists;
 use function explode;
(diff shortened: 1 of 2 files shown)


Ocramius:
@akrabat should the baseline be reduced, given your explanation? :thinking:

akrabat:
> @akrabat should the baseline be reduced, given your explanation? 🤔

@Ocramius I don't think we can as the fallback if `$value` is not a `string`, `array` or instance of `UploadedFileInterface` is:

```php
return $this->getFileBasedFileInfo($value, $hasType, $hasBasename);
```

Within `getFileBasedFileInfo()`, `$value` is treated as a string, and I'm pretty sure any object that has a `__toString()` would work.

Ocramius:
I'm pretty sure any object that has a `__toString()` would work.

Perhaps we should limit to `Stringable` then? Will try to catch you today about this 👍

src/File/ExcludeMimeType.php
```diff
@@ -36,8 +36,8 @@ class ExcludeMimeType extends MimeType
      * of mimetypes can be checked. If you give for example "image" all image
      * mime types will not be accepted like "image/gif", "image/jpeg" and so on.
      *
-     * @param  string|array|object $value Real file to check for mimetype
-     * @param  array               $file  File data from \Laminas\File\Transfer\Transfer (optional)
+     * @param  string|array|UploadedFileInterface $value Real file to check for mimetype
```
Ocramius, line 41: You wanted stringable here? This package has no knowledge of PSR-7

akrabat:
Thanks @Ocramius.

I've checked the code in a little more detail and it turns out that `$value` can only be a `string`, an `array` or an instance of `UploadedFileInterface`. It cannot be an object that's `Stringable` as we call `basename()` on which only supports native PHP strings. The only other thing to worry about is that `is_array()` is called on `$value` and again `is_array()` only works with native PHP arrays.

Updated to limit to this.

src/File/ExcludeMimeType.php
akrabat: No. Because UploadedFileInterface is not Stringable.  I should have imported.

This package already knows PSR-7 because of its current use in [FileInformationTrait](https://github.com/laminas/laminas-validator/blob/2.51.x/src/File/FileInformationTrait.php#L6).

Ocramius (approved):
LGTM, thanks @akrabat!

PR #242, opened 2024-05-14
Add `throwExceptions` option to Callback validator

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | no
| New Feature   | yes

### Description

This is a handy option for the callback validator that stops it from swallowing exceptions thrown in your callback. Off by default preserves existing behaviour.

Why? particularly in development, if you have a complex callback, can't figure out why it's failing and step debugging isn't helping you either

The Psalm baseline is mostly improved but there are still issues with the options array - this is mainly because of an `@property array $options` doc block on `AbstractValidator`

### DIFF
--- a/docs/book/validators/callback.md
+++ b/docs/book/validators/callback.md
@@ -10,6 +10,7 @@ The following options are supported for `Laminas\Validator\Callback`:
 - `callback`: Sets the callback which will be called for the validation.
 - `callbackOptions`: Sets the additional options which will be given to the validator
   and/or callback.
+- `throwExceptions`: When true, [allows exceptions thrown inside of callbacks to propagate](#exceptions-within-callbacks).
 
 ## Basic usage
 
--- a/docs/book/validators/callback.md
+++ b/docs/book/validators/callback.md
@@ -204,7 +204,7 @@ which can be used is not limited.
 
 ## Exceptions within Callbacks
 
-By default, the callback validator will catch any `Throwable` thrown inside the callback and return false.
+By default, the callback validator will catch any `Exception` thrown inside the callback and return false.
 The error message will indicate callback failure as opposed to invalid input.
 
 There is a third option `throwExceptions` that when `true` will re-throw exceptions that occur inside the callback.
--- a/docs/book/validators/callback.md
+++ b/docs/book/validators/callback.md
@@ -214,8 +214,6 @@ This is primarily useful in a development environment when you are testing callb
 For example:
 
 ```php
-use Laminas\Validator\Callback;
-
 $callback = static function (mixed $value): bool {
     if ($value === true) {
         return true;
@@ -224,7 +222,7 @@ $callback = static function (mixed $value): bool {
     throw new ApplicationException('Bad news');
 }
 
-$validator = new Callback([
+$validator = new Laminas\Validator\Callback([
     'callback' => $callback,
     'throwExceptions' => true,
 ]);
--- a/docs/book/validators/callback.md
+++ b/docs/book/validators/callback.md
@@ -201,3 +201,33 @@ When making the call to the callback, the value to be validated will always be
 passed as the first argument to the callback followed by all other values given
 to `isValid()`; all other options will follow it. The amount and type of options
 which can be used is not limited.
+
+## Exceptions within Callbacks
+
+By default, the callback validator will catch any `Throwable` thrown inside the callback and return false.
+The error message will indicate callback failure as opposed to invalid input.
+
+There is a third option `throwExceptions` that when `true` will re-throw exceptions that occur inside the callback.
+
+This is primarily useful in a development environment when you are testing callbacks and need to catch and verify exceptions thrown by your own application.
+
+For example:
+
+```php
+use Laminas\Validator\Callback;
+
+$callback = static function (mixed $value): bool {
+    if ($value === true) {
+        return true;
+    }
+    
+    throw new ApplicationException('Bad news');
+}
+
+$validator = new Callback([
+    'callback' => $callback,
+    'throwExceptions' => true,
+]);
+
+$validator->isValid('Nope'); // An exception is thrown
+```
--- a/src/Callback.php
+++ b/src/Callback.php
@@ -19,7 +19,8 @@ use function is_callable;
  *     callback: callable,
  *     callbackOptions?: array<array-key, mixed>,
  *     throwExceptions?: bool,
- * }&array<string, mixed>
+ *     ...<string, mixed>
+ * }
  */
 class Callback extends AbstractValidator
 {
--- a/src/Callback.php
+++ b/src/Callback.php
@@ -10,11 +10,16 @@ use function call_user_func_array;
 use function is_callable;
 
 /**
- * @psalm-type Options = array{
- *     callback?: callable|null,
+ * @psalm-type OptionsProperty = array{
+ *     callback: callable|null,
+ *     callbackOptions: array<array-key, mixed>,
+ *     throwExceptions: bool,
+ * }
+ * @psalm-type OptionsArgument = array{
+ *     callback: callable,
  *     callbackOptions?: array<array-key, mixed>,
  *     throwExceptions?: bool,
- * }
+ * }&array<string, mixed>
  */
 class Callback extends AbstractValidator
 {
@@ -41,7 +46,7 @@ class Callback extends AbstractValidator
     /**
      * Default options to set for the validator
      *
-     * @var Options
+     * @var OptionsProperty
      */
     protected $options = [
         'callback'        => null, // Callback in a call_user_func format, string || array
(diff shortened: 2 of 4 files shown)


src/Callback.php
```diff
@@ -139,8 +144,13 @@ public function isValid($value, $context = null)
                 $this->error(self::INVALID_VALUE);
                 return false;
             }
-        } catch (Exception) {
+        } catch (Throwable $error) {
```
Ocramius: Careful: `Throwable` used to be thrown by default :thinking:

Ocramius (approved):
LGTM, but there is a very mild BC break in `Exception` to `Throwable` behavior change.

I'm fine with this code now suppressing `Throwable` by default, but it should be noted somewhere :thinking:

src/Callback.php
gsteel: Ack. I've pushed docs for the option - will this be sufficent do you think?

cc @froschdesign

Xerkus: Should validator be swallowing php errors?

gsteel: I'd say not, but it's what we have

Xerkus: We don't have it. Yet. 
This change will make validator swallow php type errors, calling with fewer parameters or invoking callback that is not callable and such. All the subclasses of Error that I can think of are a direct result of a bug in the code.

gsteel: Ah, I see what you mean.

Is this still true considering that `strict_types` is off here?

I think it's probably worth reverting this change anyhow. Ideally, nothing would be caught at all and this whole patch would be unnecessary.

Ocramius: I think this breakage is fine BTW, just requires some sort of release notes.

docs/book/validators/callback.md
```diff
@@ -201,3 +201,33 @@ When making the call to the callback, the value to be validated will always be
 passed as the first argument to the callback followed by all other values given
 to `isValid()`; all other options will follow it. The amount and type of options
 which can be used is not limited.
+
+## Exceptions within Callbacks
```
froschdesign, line 206: The new option must also be listed at the top of the page:

https://github.com/laminas/laminas-validator/blob/405c711a7a6b75cc00a906bf756324a2599a2ff5/docs/book/validators/callback.md?plain=1#L6-L12

A link to this section would be great.

docs/book/validators/callback.md
```diff
@@ -201,3 +201,33 @@ When making the call to the callback, the value to be validated will always be
...
+        return true;
+    }
+    
+    throw new ApplicationException('Bad news');
+}
+
+$validator = new Callback([
+    'callback' => $callback,
+    'throwExceptions' => true,
+]);
```
froschdesign: We do not yet have a fixed definition for this but the class with namespace is only used once and all other code examples follow this way:

```suggestion
```php
$callback = static function (mixed $value): bool {
    if ($value === true) {
        return true;
    }
    
    throw new ApplicationException('Bad news');
}

$validator = new Laminas\Validator\Callback([
    'callback' => $callback,
    'throwExceptions' => true,
]);
```

froschdesign (changes requested):
It's always nice to see that the documentation is also included. 👍🏻 

I have added two small suggestions.

src/Callback.php
gsteel: I've reverted back to `Exception`, you never know, someone out there might be catching `Error` 🤪

With your blessing @Xerkus I'll merge and release.

froschdesign (commented):
LGTM 👍🏻 

I also like the new Psalm type for the options!

src/Callback.php
Xerkus: > Is this still true considering that `strict_types` is off here?

Huh. Unexpected.

Anyway, that leaves catchable errors coming from callback call stack.
I would add inline comment for later however. Something like "Intentionally excluding catchable errors as they are indicative of a bug and should not be suppressed"

My reasoning for allowing errors through is mostly because of user experience. Silly non-fatal mistake in the code can lead to confusing validation failure of a form or when tinkering with input filters. Debugging input filter specification can be daunting even without the error getting suppressed.

src/Callback.php
```diff
@@ -9,6 +9,13 @@
 use function call_user_func_array;
 use function is_callable;
 
+/**
+ * @psalm-type Options = array{
+ *     callback?: callable|null,
```
Xerkus: This is wrong for `options` property and it is wrong for `setOptions()` or constructor.
For options property all 3 keys are always defined.
For `setOptions()` this must include AbstractOptions shape from the parent.

src/Callback.php
```diff
@@ -34,18 +41,15 @@ class Callback extends AbstractValidator
     /**
      * Default options to set for the validator
      *
-     * @var mixed
+     * @var Options
```
Xerkus: This is _not_ the same as options passed to `setOptions()` or constructor.

src/Callback.php
```diff
@@ -34,18 +41,15 @@ class Callback extends AbstractValidator
...
         'callbackOptions' => [], // Options for the callback
+        'throwExceptions' => false, // Whether to throw exceptions raised within the callback or not
     ];
 
-    /**
-     * Constructor
-     *
-     * @param array|callable $options
-     */
+    /** @param Options|callable $options */
```
Xerkus: options here also include generic validator options, so this type info is not entirely correct. See above comment.

src/Callback.php
```diff
@@ -58,17 +62,17 @@ public function __construct($options = null)
     /**
      * Returns the set callback
      *
-     * @return mixed
+     * @return callable|null
      */
     public function getCallback()
     {
-        return $this->options['callback'];
+        return $this->options['callback'] ?? null;
```
Xerkus: This change is required because of wrong type declaration for a property.

src/Callback.php
```diff
@@ -85,16 +89,17 @@ public function setCallback($callback)
     /**
      * Returns the set options for the callback
      *
-     * @return mixed
+     * @return array<array-key, mixed>
      */
     public function getCallbackOptions()
     {
-        return $this->options['callbackOptions'];
+        return $this->options['callbackOptions'] ?? [];
```
Xerkus: Same as above

src/Callback.php
```diff
@@ -139,8 +144,13 @@ public function isValid($value, $context = null)
                 $this->error(self::INVALID_VALUE);
                 return false;
             }
-        } catch (Exception) {
+        } catch (Exception $error) {
```
Xerkus: Minor readability nitpick: rename `$exception` or `$e` as it does not include errors. Add comment for future reference

Xerkus (changes requested):
Psalm type declaration needs adjusting

src/Callback.php
gsteel: Including all of the abstract options cannot be done without implementing templates thoughout the entire codebase. Psalm does not support intersections on imported types:
https://psalm.dev/r/be51b8b478

Separate shapes for the property vs the argument, fine, but I'm not going to do the work required to get the intersection with AbstractOptions working.

src/Callback.php
gsteel: See https://github.com/laminas/laminas-validator/pull/242#discussion_r1603117308

src/Callback.php
Xerkus: It is an array shape iirc, so something like this should work: https://psalm.dev/r/4b68c89299

Edit: nevermind, psalm loses the shape

gsteel: I defined the shape on AbstractValidator - It can't be added as an intersection via `@psalm-import-type AbstractOptions from AbstractValidator` because psalm yields `InvalidDocblock` - I have added an intersection with `array<string, mixed>` which is as far as I'm willing to go with this.

src/Callback.php
```diff
@@ -9,6 +9,18 @@
...
+ * @psalm-type OptionsProperty = array{
+ *     callback: callable|null,
+ *     callbackOptions: array<array-key, mixed>,
+ *     throwExceptions: bool,
+ * }
+ * @psalm-type OptionsArgument = array{
+ *     callback: callable,
+ *     callbackOptions?: array<array-key, mixed>,
+ *     throwExceptions?: bool,
+ * }&array<string, mixed>
```
Xerkus: property does not need separate type declared imo.

Wouldn't it be preferred to copy options from the parent?

Psalm has unsealed array syntax:
```suggestion
 * @psalm-type OptionsArgument = array{
 *     callback: callable,
 *     callbackOptions?: array<array-key, mixed>,
 *     throwExceptions?: bool,
 *     ...<string, mixed>
 * }
```

src/Callback.php
Xerkus: I thought psalm supported array unpacking syntax for shapes. Apparently it does not, it is only used for unsealed array feature.

src/Callback.php
gsteel: The reason I used another type was to make all options optional except for the callable. i.e. `throwExceptions?: bool` instead of `throwExceptions: bool`

Xerkus (approved):
Unsealed array for options works.

Copying declaration from parent is a maintenance burden for later and changing validators to extend psalm templates is too much effort for this PR.

Xerkus:
LGTM

PR #245, opened 2024-06-05
Remove obsolete PHP 8.1 shims

derrabus:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

This PR removes test shims introduced by #107. All the affected packages are required in PHP 8.1 compatible versions nowadays, so all of those shims should be obsolete by now. Let's test against the real thing!


gsteel (approved):
Thank you @derrabus

PR #249, opened 2024-06-12
Remove deprecated Laminas\Db validators

gsteel:
Needs #248 first


mano87:
@gsteel @weierophinney 
That's great when the validator is removed. But we use it in many places. What should we do now? Why remove it without having an alternative?

PR #252, opened 2024-06-12
Enable strict types throughout

gsteel:
Also fixes a couple of bugs surfaced by turning strict types on.

It's worth reviewing commits in isolation due to the number of changed files…

### DIFF
--- a/src/File/Size.php
+++ b/src/File/Size.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Laminas\Validator\File;
 
-use Laminas\Stdlib\ErrorHandler;
 use Laminas\Validator\AbstractValidator;
 use Laminas\Validator\Exception;
 use Traversable;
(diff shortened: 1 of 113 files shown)


src/File/Size.php
```diff
@@ -251,7 +254,7 @@ public function isValid($value, $file = null)
 
         // limited to 4GB files
         ErrorHandler::start();
-        $size = sprintf('%u', filesize($fileInfo['file']));
+        $size = (int) sprintf('%u', filesize($fileInfo['file']));
```
froschdesign: Looks like a duplication: type casting and `sprintf` with `%u`?!

gsteel: Yes, but psalm does not recognise the coercion:

https://psalm.dev/r/0fd48ff1ba

froschdesign: And killing `sprintf` is not a solution here?

gsteel: Good point! The error handler can get in the bin too 😂

froschdesign (approved):

PR #254, opened 2024-06-12
Remove dependency on laminas-filter

gsteel:
This dependency was always pretty unnecessary. It will make the digits validator usable without installing an optional dependency and make migration to SMv4 easier too.


froschdesign (approved):
Go for it! 👍🏻

PR #255, opened 2024-06-12
Mark all classes as soft `@final` where possible

gsteel:
To signal intent to make all validators final in v3, this patch adds `@final` to all classes where possible.

A number of suspect documented types have been altered because psalm found them…

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -16,14 +16,6 @@
     </MixedArgument>
   </file>
   <file src="src/AbstractValidator.php">
-    <LessSpecificReturnStatement>
-      <code><![CDATA[$this]]></code>
-      <code><![CDATA[$this]]></code>
-      <code><![CDATA[$this]]></code>
-      <code><![CDATA[$this]]></code>
-      <code><![CDATA[$this]]></code>
-      <code><![CDATA[$this]]></code>
-    </LessSpecificReturnStatement>
     <MixedArgumentTypeCoercion>
       <code><![CDATA[$value]]></code>
     </MixedArgumentTypeCoercion>
@@ -367,15 +359,6 @@
     </PossiblyUnusedMethod>
   </file>
   <file src="src/Bitwise.php">
-    <DocblockTypeContradiction>
-      <code><![CDATA[self::OP_AND === $this->operator]]></code>
-      <code><![CDATA[self::OP_AND === $this->operator]]></code>
-      <code><![CDATA[self::OP_XOR === $this->operator]]></code>
-      <code><![CDATA[self::OP_XOR === $this->operator]]></code>
-    </DocblockTypeContradiction>
-    <InvalidPropertyAssignmentValue>
-      <code><![CDATA[$operator]]></code>
-    </InvalidPropertyAssignmentValue>
     <MixedAssignment>
       <code><![CDATA[$temp['control']]]></code>
       <code><![CDATA[$temp['operator']]]></code>
@@ -474,6 +457,9 @@
       <code><![CDATA[is_array($options)]]></code>
       <code><![CDATA[null === $this->session]]></code>
     </DocblockTypeContradiction>
+    <ImpureByReferenceAssignment>
+      <code><![CDATA[$session->tokenList]]></code>
+    </ImpureByReferenceAssignment>
     <MixedArgument>
       <code><![CDATA[$key]]></code>
       <code><![CDATA[$session->tokenList[$tokenId]]]></code>
@@ -1657,9 +1643,6 @@
     </RedundantCastGivenDocblockType>
   </file>
   <file src="src/Ip.php">
-    <LessSpecificImplementedReturnType>
-      <code><![CDATA[AbstractValidator]]></code>
-    </LessSpecificImplementedReturnType>
     <MixedArgument>
       <code><![CDATA[$regex]]></code>
       <code><![CDATA[$regex]]></code>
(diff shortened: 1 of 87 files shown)


src/Uuid.php
```diff
@@ -7,6 +7,8 @@
 
 /**
  * Uuid validator.
+ *
+ * @final
```
froschdesign: This class is already marked as final.

src/GpsPoint.php
```diff
@@ -10,6 +10,7 @@
 use function str_contains;
 use function str_replace;
 
+/** @final */
```
froschdesign: This class is already marked as final.

src/IsArray.php
```diff
@@ -7,6 +7,7 @@
 use function get_debug_type;
 use function is_array;
 
+/** @final */
```
froschdesign: This class is already marked as final.

src/IsJsonString.php
```diff
@@ -24,6 +24,7 @@
  * }
  * @psalm-import-type AbstractOptions from AbstractValidator
  * @psalm-type Options = AbstractOptions|CustomOptions
+ * @final
```
froschdesign: This class is already marked as final.

src/UndisclosedPassword.php
```diff
@@ -15,6 +15,7 @@
 use function strtoupper;
 use function substr;
 
+/** @final */
```
froschdesign: This class is already marked as final.

froschdesign (changes requested):

gsteel:
Thanks @froschdesign 🙏

froschdesign (approved):
🚢 it!

InvisibleSmiley:
Some people (like us) extend validators that are now "final" with the intent of either changing the validation failure message (e.g. `InArray::$messageTemplates` to include `%value%` in the message) or creating otherwise pre-configured validators (e.g. by passing fixed values for the `haystack` and `strict` options in an `InArray` constructor override).

The former can alternatively be achieved by passing the `messageTemplates` option to the constructor or calling `setMessage`, but what is the recommended alternative for creating pre-configured validators? I need something that I can reference via FQCN (`::class`) in the end.

AFAICS there is no "abstract validator proxy class" readily available, and I don't want to extend `AbstractValidator` or implement `ValidatorInterface` myself multiple times.

I hope you can clear things up, at the latest when you write the v3 migration.
Until I have a viable alternative for my use case I'll just ignore the added `@final`.

froschdesign:
@InvisibleSmiley 
> …creating otherwise pre-configured validators…

The options are the key point. Here is one example for an input filter:

```php
[
    'name'       => 'title',
    'required'   => true,
    'filters'    => [
        // …
    ],
    'validators' => [
        [
            'name'    => Laminas\Validator\InArray::class,
            'options' => Example\Validator\Options::INARRAY_FIXED, // Constant in a custom class as an array
        ],
    ],
],
```

gsteel:
@InvisibleSmiley There are a few ways to handle this, for example composing a validator inside another:

```php
declare(strict_types=1);

namespace LaminasTest\Validator;

use Laminas\ServiceManager\Factory\InvokableFactory;
use Laminas\Validator\InArray;
use Laminas\Validator\ValidatorInterface;

final class ExampleComposedValidator implements ValidatorInterface
{
    private readonly InArray $validator;

    public function __construct()
    {
        $this->validator = new InArray([
            'strict'           => InArray::COMPARE_STRICT,
            'messageTemplates' => [
                InArray::NOT_IN_ARRAY => '%value% not found…',
            ],
        ]);
    }

    public function isValid(mixed $value): bool
    {
        return $this->validator->isValid($value);
    }

    public function getMessages(): array
    {
        return $this->validator->getMessages();
    }
}

$config = [
    'validators' => [
        'factories' => [
            ExampleComposedValidator::class => InvokableFactory::class,
        ],
    ],
];
```

Creating a marker interface and setting up a factory to return whatever validator you need:

```php
interface FirstNameValidator extends ValidatorInterface
{
}

final class FirstNameValidatorFactory
{
    public function __invoke(ContainerInterface $container): ValidatorInterface
    {
        $chain = new ValidatorChain();
        $chain->attachByName(
            NotEmpty::class,
            [
                'messages' => [
                    NotEmpty::IS_EMPTY => 'Oh no!',
                ],
            ],
            true,
        );

        $chain->attachByName(
            Regex::class,
            [
                'pattern' => [
                    '/^Fred/',
                ],
            ],
        );

        return $chain;
    }
}

$config = [
    'validators' => [
        'factories' => [
            FirstNameValidator::class => FirstNameValidatorFactory::class,
        ],
    ],
];
```

You could of course override default implementations with your own factories, such as

```php
$config = [
    'validators' => [
        'factories' => [
            \Laminas\Validator\InArray::class => SomeFactory::class,
        ],
    ],
];
```

Personally, I don't tend to use validator stand-alone, but instead create concrete `InputFilter` instances. For re-usable input specifications, I write invokable classes to represent things in the domain that have specific validation needs such as:

```php
/** @psalm-import-type InputSpecification from InputFilterInterface */
class IataAirportCodeInputSpec
{
    /** @return InputSpecification */
    public function __invoke(bool $required = true): array
    {
        $validators = $required ? [
            'notEmpty' => [
                'name' => NotEmpty::class,
                'options' => [
                    'messages' => [ NotEmpty::IS_EMPTY => 'Please provide an IATA Airport Code' ],
                ],
            ],
            'isCode' => [ 'name' => IataAirportCode::class ],
        ] : [
            'isCode' => [ 'name' => IataAirportCode::class ],
        ];

        return [
            'required' => $required,
            'validators' => $validators,
            'filters' => [
                'trim' => [ 'name' => StringTrim::class ],
                'upper' => [ 'name' => StringToUpper::class ],
                'toNull' => [ 'name' => ToNull::class ],
                'toCode' => [ 'name' => ToIataAirportCode::class ],
            ],
        ];
    }
}

final class AirportInputFilter extends InputFilter
{
    public function init(): void
    {
        $this->add((new IataAirportCodeInputSpec())(true), 'code');
        // ... other inputs ...
    }
}
```

InvisibleSmiley:
Thanks for the pointers.

I went with the composition approach but such that constructor (parameters) and constants stay compatible with the original. That way it's less of a breaking change.

```php
<?php

declare(strict_types=1);

use Laminas\Stdlib\ArrayUtils;
use Laminas\Validator\InArray;
use Laminas\Validator\ValidatorInterface;
use Traversable;

class MyInArray implements ValidatorInterface
{
    public final const NOT_IN_ARRAY = 'notInArray';

    public final const COMPARE_STRICT = InArray::COMPARE_STRICT;
    public final const COMPARE_NOT_STRICT = InArray::COMPARE_NOT_STRICT;
    public final const COMPARE_NOT_STRICT_AND_PREVENT_STR_TO_INT_VULNERABILITY = InArray::COMPARE_NOT_STRICT_AND_PREVENT_STR_TO_INT_VULNERABILITY;

    private readonly InArray $validator;

    /**
     * @param iterable<string, mixed>|null $options
     */
    public function __construct(?iterable $options = null)
    {
        if ($options instanceof Traversable) {
            $options = ArrayUtils::iteratorToArray($options);
        }

        if (!isset($options['messageTemplates'][self::NOT_IN_ARRAY])) {
            $options['messageTemplates'][self::NOT_IN_ARRAY] = 'The input "%value%" was not found in the haystack';
        }

        $this->validator = new InArray($options);
    }

    public function isValid(mixed $value): bool {
        return $this->validator->isValid($value);
    }

    public function getMessages(): array {
        return $this->validator->getMessages();
    }
}
```

PR #257, opened 2024-06-14
Add new Validator: `HostWithPublicIPv4Address`

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

Whilst doing some refactoring for v3, I found that the Email Address validator validates IP addresses/hostnames for reserved IPs.

In order to test this effectively, it makes more sense to implement this as a validator and compose that into the Email address validator - it is likely to prove useful to some users too in a standalone context.

So, this validator is a pretty straightforward extraction of https://github.com/laminas/laminas-validator/blob/2.57.x/src/EmailAddress.php#L282-L344

The expectation is that the functionality in the Email validator will be replaced with this.

### DIFF
--- a/src/HostWithPublicIPv4Address.php
+++ b/src/HostWithPublicIPv4Address.php
@@ -4,14 +4,60 @@ declare(strict_types=1);
 
 namespace Laminas\Validator;
 
+use function explode;
+use function filter_var;
 use function get_debug_type;
 use function gethostbynamel;
+use function ip2long;
 use function is_array;
 use function is_string;
-use function preg_match;
+use function pow;
+
+use const FILTER_FLAG_GLOBAL_RANGE;
+use const FILTER_FLAG_IPV4;
+use const FILTER_FLAG_NO_PRIV_RANGE;
+use const FILTER_FLAG_NO_RES_RANGE;
+use const FILTER_VALIDATE_IP;
+use const PHP_VERSION_ID;
 
 final class HostWithPublicIPv4Address extends AbstractValidator
 {
+    /**
+     * Reserved CIDRs are extracted from IANA with additions from Wikipedia
+     *
+     * @link https://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml
+     * @link https://en.wikipedia.org/wiki/Reserved_IP_addresses
+     */
+    private const RESERVED_CIDR = [
+        '0.0.0.0/8',
+        '0.0.0.0/32',
+        '10.0.0.0/8',
+        '100.64.0.0/10',
+        '127.0.0.0/8',
+        '169.254.0.0/16',
+        '172.16.0.0/12',
+        '192.0.0.0/24',
+        '192.0.0.0/29',
+        '192.0.0.8/32',
+        '192.0.0.9/32',
+        '192.0.0.10/32',
+        '192.0.0.170/32',
+        '192.0.0.171/32',
+        '192.0.2.0/24',
+        '192.31.196.0/24',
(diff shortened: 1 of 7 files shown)


src/HostWithPublicIPv4Address.php
```diff
@@ -0,0 +1,98 @@
...
+        if (! is_array($addressList)) {
+            $this->error(self::ERROR_HOSTNAME_NOT_RESOLVED);
+
+            return false;
+        }
+
+        $privateAddressWasFound = false;
+
+        // phpcs:disable Generic.Files.LineLength
+        foreach ($addressList as $server) {
```
boesing, line 115: any reason why not using `filter_var` here with `FILTER_VALIDATE_IP` and `FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE `?

That would actually prevent us from having to maintain this here, especially since we do not need all these regex groups declared in those regexes here.

gsteel: 👌 Didn't know filter had those flags… Gimme a sec :)

gsteel: These filter flags let through https://datatracker.ietf.org/doc/html/rfc6598#section-7 addresses amongst others.

Even including `FILTER_FLAG_GLOBAL_RANGE` available in PHP 8.2 lets through reserved IPs

:(

src/HostWithPublicIPv4Address.php
```diff
@@ -0,0 +1,98 @@
...
+
+        if (! is_string($value)) {
+            $this->error(self::ERROR_NOT_STRING);
+
+            return false;
+        }
+
+        $this->value = $value;
+
+        if (! preg_match('/^([0-9]{1,3}\.){3}[0-9]{1,3}$/', $value)) {
```
Ocramius: This kinda allows `999.999.999.999` :|

src/HostWithPublicIPv4Address.php
```diff
@@ -0,0 +1,98 @@
...
+            $this->error(self::ERROR_HOSTNAME_NOT_RESOLVED);
+
+            return false;
+        }
+
+        $privateAddressWasFound = false;
+
+        // phpcs:disable Generic.Files.LineLength
+        foreach ($addressList as $server) {
+            // Search for 0.0.0.0/8, 10.0.0.0/8, 127.0.0.0/8
```
Ocramius: Many of the below are reserved blocks: worth linking to where you found these blocks in the RFCs

src/HostWithPublicIPv4Address.php
gsteel: I confess, this is just copied from EmailAddress validator with the assumption that it was correct.

Looks like I'll have to figure this out from https://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml

from scratch to ensure it's correct.

src/HostWithPublicIPv4Address.php
boesing: Can we then partly use the `filter_var` and if that is failing, apply just regexes for those which we know are not validated by `*_NO_RES_RANGE`?

Not a real requirement for me, just want to simplify this a bit. Was not aware that this is more or less copy & paste from the email thingy.

boesing (approved):
LGTM, still a bunch of manually listed IPs.
Was not expecting that there are that many ranges which are not covered by the `filter_var` stuff. Good to know.

PR #262, opened 2024-06-16
Refactor URI Validator, Drop dependency on Laminas\Uri

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| QA            | yes

### Description

Refactors the URI validator to use `parse_url` internally instead of `Laminas\Uri` which is in security-only mode. URL validation is likely more lenient now but due to the lack of existing tests, it's difficult to say how much of an impact this might have.

Also refactors the `Sitemap\Loc` validator - this validator has been improved to check the max length of the URI according to the sitemap spec, and now rejects URIs that are not correctly URL encoded - i.e. they'd cause problems in an XML file.

### DIFF
--- a/src/Sitemap/Loc.php
+++ b/src/Sitemap/Loc.php
@@ -4,57 +4,92 @@ declare(strict_types=1);
 
 namespace Laminas\Validator\Sitemap;
 
-use Laminas\Uri;
 use Laminas\Validator\AbstractValidator;
+use Laminas\Validator\Uri;
 
+use function html_entity_decode;
+use function htmlentities;
 use function is_string;
+use function str_replace;
+use function strlen;
 
 /**
  * Validates whether a given value is valid as a sitemap <loc> value
  *
- * @link       http://www.sitemaps.org/protocol.php Sitemaps XML format
- * @see        Laminas\Uri\Uri
+ * @link https://www.sitemaps.org/protocol.html Sitemaps XML format
  */
 final class Loc extends AbstractValidator
 {
-    /**
-     * Validation key for not valid
-     */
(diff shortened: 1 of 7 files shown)


src/Sitemap/Loc.php
```diff
@@ -4,59 +4,92 @@
...
- * @final
+ * @link https://www.sitemaps.org/protocol.html Sitemaps XML format
  */
-class Loc extends AbstractValidator
+final class Loc extends AbstractValidator
 {
-    /**
-     * Validation key for not valid
-     */
+    private const URI_MAX_LENGTH = 2048;
```
Ocramius, line 23: hmm, 2048? Really this short?

gsteel: Yep - that's what it says at https://www.sitemaps.org/protocol.html#locdef 🤷‍♂️

Ocramius (approved):
LGTM: nice dependency removal there :-)

Ocramius:
needs a rebase D:

src/Sitemap/Loc.php
```diff
@@ -4,59 +4,92 @@
...
+use function strlen;
 
 /**
  * Validates whether a given value is valid as a sitemap <loc> value
  *
- * @link       http://www.sitemaps.org/protocol.php Sitemaps XML format
- * @see        Laminas\Uri\Uri
- *
- * @final
+ * @link https://www.sitemaps.org/protocol.html Sitemaps XML format
```
froschdesign, line 19: 👍🏻

froschdesign (approved):

gsteel:
Thanks for the reviews @Ocramius & @froschdesign 👍

PR #267, opened 2024-06-17
Upgrade to Service Manager v4

### LINKED ISSUE #131: Drop Service Manager v2 Compatibility
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes/no
| BC Break    | yes/no

#### Summary

with #129, minimum required version of service manager is 3.12.0, so we can scan for 2.x related compatibility code and remove it.

### LINKED ISSUE #206: Need support for laminas/laminas-servicemanager v4
laminas/laminas-validator cannot be installed with laminas/laminas-servicemanager v4, that support psr/container v2.
Need support.

gsteel:
|    Q          |   A
|-------------- | -----
| BC Break      | yes
| New Feature   | yes

### Description

Initial support for Service Manager v4

- [x] Remove the shipped CSRF validator
- [x] Remove dependency on `laminas-session`
- [x] Support Service Manager v4 in `laminas-i18n`, or swap the dependency with the proposed [laminas-translator component](https://github.com/gsteel/laminas-translator)
- [x] Remove temporary dev branch for `i18n` in `composer.json`

Related https://github.com/laminas/laminas-session/pull/86

- Closes #233

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
     },
     "require": {
         "php": "~8.1.0 || ~8.2.0 || ~8.3.0",
-        "laminas/laminas-servicemanager": "^3.21.0",
+        "laminas/laminas-servicemanager": "^4.1.0",
         "laminas/laminas-stdlib": "^3.13",
         "laminas/laminas-translator": "^1.0",
         "psr/http-message": "^1.0.1 || ^2.0.0"
--- a/src/ValidatorChain.php
+++ b/src/ValidatorChain.php
@@ -105,7 +105,7 @@ final class ValidatorChain implements Countable, IteratorAggregate, ValidatorInt
     public function plugin($name, ?array $options = null)
     {
         $plugins = $this->getPluginManager();
-        return $plugins->get($name, $options);
+        return $plugins->build($name, $options);
     }
 
     /**
--- a/src/ValidatorPluginManager.php
+++ b/src/ValidatorPluginManager.php
@@ -246,7 +246,7 @@ final class ValidatorPluginManager extends AbstractSingleInstancePluginManager
 
         if ($container->has('MvcTranslator')) {
             $translator = $container->get('MvcTranslator');
-            assert($translator instanceof Translator\TranslatorInterface);
+            assert($translator instanceof TranslatorInterface);
             $validator->setTranslator($translator);
 
             return;
--- a/src/ValidatorPluginManager.php
+++ b/src/ValidatorPluginManager.php
@@ -4,337 +4,250 @@ declare(strict_types=1);
 
 namespace Laminas\Validator;
 
-use Laminas\ServiceManager\AbstractPluginManager;
-use Laminas\ServiceManager\Exception\InvalidServiceException;
+use Closure;
+use Laminas\ServiceManager\AbstractSingleInstancePluginManager;
 use Laminas\ServiceManager\Factory\InvokableFactory;
 use Laminas\ServiceManager\ServiceManager;
 use Laminas\Translator\TranslatorInterface;
 use Psr\Container\ContainerInterface;
 
-use function get_debug_type;
-use function method_exists;
-use function sprintf;
+use function array_replace_recursive;
+use function assert;
 
 /**
  * @psalm-import-type ServiceManagerConfiguration from ServiceManager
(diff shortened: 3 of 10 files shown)


composer.json
```diff
@@ -32,14 +32,13 @@
...
-        "laminas/laminas-servicemanager": "^3.21.0",
+        "laminas/laminas-servicemanager": "^4.1.0",
         "laminas/laminas-stdlib": "^3.13",
         "psr/http-message": "^1.0.1 || ^2.0.0"
     },
     "require-dev": {
         "laminas/laminas-coding-standard": "^2.5",
-        "laminas/laminas-i18n": "^2.26.0",
-        "laminas/laminas-session": "^2.20",
+        "laminas/laminas-i18n": "dev-relax-sm-constraint",
```
gsteel: Dropping session dependency here just so we can install SMv4 - I'll need to remove this dep in another patch before this can be merged.

We'll also need a new major of i18n

gsteel: Session dependency is solved via #270

gsteel:
Needs #301 and a rebase

src/ValidatorPluginManager.php
```diff
@@ -4,337 +4,250 @@
...
-        try {
-            $this->validate($plugin);
-        } catch (InvalidServiceException $e) {
-            throw new Exception\RuntimeException(sprintf(
-                'Plugin of type %s is invalid; must implement %s',
-                get_debug_type($plugin),
-                ValidatorInterface::class
-            ), $e->getCode(), $e);
-        }
+        $this->addInitializer(Closure::fromCallable([$this, 'injectTranslator']));
```
froschdesign, line 236: The method is deprecated, should `configure` be used?

```php
@deprecated Please use {@see AbstractPluginManager::configure()} instead.
```

https://github.com/laminas/laminas-servicemanager/blob/ae01570ecdf4559aef2f15d4e79e8c36874f7416/src/AbstractPluginManager.php#L313-L323

gsteel: Because `injectValidatorPluginManager` needs `$this`, I can't make it static, which it would need to be in order to be added to the config constant.

Not sure what to do here. Maybe deprecating `injectValidatorPluginManager` is the way to go - if a validator depends on other validators, it should have its own factory…

🤔

gsteel: On second thoughts, I could just fetch `$this` from the container because at the point of use, the plugin manager will already be up and running so there won't be a circular dependency. The problem though is multiple plugin managers in user projects which whilst weird, is still possible.

froschdesign: We can leave everything as it is for now and then rewrite it later, possibly with the help of laminas-servicemanager-migration.

froschdesign (approved):
LGTM 👍🏻

gsteel:
Thanks @froschdesign :)

froschdesign:
Sorry for the delay, I had to look at the `CommonPluginManagerTrait` of laminas-servicemanager.

src/ValidatorChain.php
```diff
@@ -105,7 +105,7 @@ public function setPluginManager(ValidatorPluginManager $plugins)
     public function plugin($name, ?array $options = null)
     {
         $plugins = $this->getPluginManager();
-        return $plugins->get($name, $options);
+        return $plugins->build($name, $options);
```
Ocramius, line 108: Is the switch to `build` intentional?

Ocramius (commented):
From my PoV, 🚢

src/ValidatorChain.php
gsteel: Yes - SMv4 no longer accepts `$options` to `get()` 👍

gsteel:
Thanks for the reviews @froschdesign and @Ocramius :)

src/ValidatorChain.php
boesing: *afair* v3 had the options only available in the `AbstractPluginManager`, but can't remember for sure.
It internally checked for `$options === null` and called either `build` with - or parent `get` without - `options`.

Not sure if this of interest here, I guess validators are not shared anyways and thus makes no diff.

PR #270, opened 2024-06-19
Remove the deprecated CSRF validator

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes

### Description

Removes the deprecated CSRF validator and the dependency on Laminas\Session


Ocramius (approved):
Besides a baseline issue, :ship:

mkdocs.yml
```diff
@@ -154,7 +153,7 @@ plugins:
             validators/bic.md: v3/validators/bic.md
             validators/callback.md: v3/validators/callback.md
             validators/credit-card.md: v3/validators/credit-card.md
-            validators/csrf.md: v3/validators/csrf.md
+            validators/csrf.md: v2/validators/csrf.md
```
froschdesign, line 156: Should be removed because the content of version 2 with the deprecation notice makes no sense for version 3.

The removal must be listed in the migration guide.

gsteel: Don't we want historical redirects to continue working through version changes?

froschdesign: Sorry, these are redirects and not navigation entries. Please ignore my comment. 🤦🏻‍♂️

PR #271, opened 2024-06-19
Clean up IP validator

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| QA            | yes

### Description

- Makes protected methods private
- Moves option handling to the constructor, allowing only `array` (Moving away from `setOptions`)
- General cleanup
- Update tests so that options used per test are easier to understand/read


froschdesign (approved):
LGTM 👍🏻

PR #272, opened 2024-06-19
Refactor `LessThan` validator

gsteel:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes
| BC Break      | yes
| QA            | yes

### Description

- Makes the `max` options actually required, fixing a bug where `null` would be the implicit max instead of throwing the expected exception
- Only allows a regular array as the options argument

### DIFF
--- a/src/LessThan.php
+++ b/src/LessThan.php
@@ -4,144 +4,85 @@ declare(strict_types=1);
 
 namespace Laminas\Validator;
 
-use Laminas\Stdlib\ArrayUtils;
-use Traversable;
-
 use function array_key_exists;
-use function array_shift;
-use function func_get_args;
-use function is_array;
+use function is_numeric;
 
+/**
+ * @psalm-type OptionsArgument = array{
+ *     max: numeric,
+ *     inclusive?: bool,
+ *     ...<string, mixed>,
+ * }
+ */
 final class LessThan extends AbstractValidator
 {
     public const NOT_LESS           = 'notLessThan';
     public const NOT_LESS_INCLUSIVE = 'notLessThanInclusive';
-
+    public const NOT_NUMERIC        = 'notNumeric';
     /**
      * Validation failure message template definitions
      *
-     * @var array
+     * @var array<string, string>
      */
-    protected $messageTemplates = [
+    protected array $messageTemplates = [
         self::NOT_LESS           => "The input is not less than '%max%'",
         self::NOT_LESS_INCLUSIVE => "The input is not less or equal than '%max%'",
+        self::NOT_NUMERIC        => "Expected a numeric value",
     ];
 
     /**
      * Additional variables available for validation failure messages
      *
-     * @var array
+     * @var array<string, string>
      */
-    protected $messageVariables = [
+    protected array $messageVariables = [
         'max' => 'max',
     ];
 
     /**
      * Maximum value
-     *
-     * @var mixed
      */
-    protected $max;
+    protected float $max;
 
     /**
      * Whether to do inclusive comparisons, allowing equivalence to max
      *
      * If false, then strict comparisons are done, and the value may equal
      * the max option
-     *
-     * @var bool
      */
-    protected $inclusive;
+    private bool $inclusive;
 
     /**
      * Sets validator options
      *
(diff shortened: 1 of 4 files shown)


src/LessThan.php
```diff
@@ -20,126 +21,58 @@ final class LessThan extends AbstractValidator
...
         'max' => 'max',
     ];
 
     /**
      * Maximum value
-     *
-     * @var mixed
      */
-    protected $max;
+    protected int|string|float $max;
```
froschdesign: The documentation says:

> `Laminas\Validator\LessThan` supports only the validation of numbers. Strings or dates can not be validated with this validator.

Why is a string allowed?

```php
$validator = new Laminas\Validator\LessThan(['max' => 'test']);

var_dump($validator->isValid(2)); // bool(true)
var_dump($validator->isValid('test')); // bool(false)
```
🤷🏻‍♂️

gsteel: Nice. I hadn't looked at the docs - the tests are doing loads of alpha comparisons so I assumed it was all part of the plan even though it was a bit 🤢

froschdesign:
@gsteel 
Found another / bigger problem because the validators `LessThan` and `GreaterThan` are used in the `DateTime` form elements:

```php
        if (
            isset($this->attributes['min'])
            && $this->valueIsValidDateTimeFormat((string) $this->attributes['min'])
        ) {
            $validators[] = new GreaterThanValidator([
                'min'       => $this->attributes['min'],
                'inclusive' => true,
            ]);
```

https://github.com/laminas/laminas-form/blob/8e3cc77c54774b9b542f52b4df5953411e0a7ab7/src/Element/AbstractDateTime.php#L107-L114

```php
        if (
            isset($this->attributes['max'])
            && $this->valueIsValidDateTimeFormat((string) $this->attributes['max'])
        ) {
            $validators[] = new LessThanValidator([
                'max'       => $this->attributes['max'],
                'inclusive' => true,
            ]);
```


https://github.com/laminas/laminas-form/blob/8e3cc77c54774b9b542f52b4df5953411e0a7ab7/src/Element/AbstractDateTime.php#L127-L134

gsteel:
@froschdesign 

Maybe we should introduce new `DateComparison` and `NumberComparison` validators in the 2.x series and deprecate `LessThan`, `GreaterThan` and `Between` for removal in 3.0

froschdesign:
@gsteel 
> …`DateComparison` and `NumberComparison`…

This would make the use clearer and would also not lead to incorrect use as in the form elements.

boesing:
but please do not add too many options, id rather have numberlessthan and Datelessthan, etc. than having a Single one which has 10 options to have it open/closed in a specific direction.

gsteel:
@boesing - 3 options. min, max and inclusive. The date validator will also need `inputFormat` to accept input dates as strings

boesing:
> [...] The date validator will also need `inputFormat` to accept input dates as strings

But a date validator could also consume `DateTimeImmutable` 🤷🏼‍♂️ 

I am still not 100% sure if first filters are applied prior validation (in `laminas-form` context) or the other way around. Still confuses me as I never understood the logic cuz it was not the way I was expecting.

So in case filters are applied prior validation, having a date validator requiring a date format would be kinda - cluttered.

---

> 3 options. min, max and inclusive

I'd rather have two validators which both consume `inclusive` with useful names than having a all-in-one validator. When checking out forms and their validators, its more understandable from the validator names than from scanning options of validators. Thats why I would always prefer more validators over "one validator covers it all". But in the end, up 2 you. Just dropping my thoughts here.

gsteel:
In laminas form, it's always filter then validate unless it's an uploaded file, then it's the other way around.

Any date comparison validator created will need to be able to validate both date instances and string input because it's not possible to know what kind of filters the user might have setup.

It should actually be _easier_ to use a number validator with `min`, `max` and inclusive (defaulted to true), because it maps really nicely onto HTML attributes of number and text inputs - easier than deciding which validator to use based on `<input type="number" min="5" max="10">`

gsteel:
This validator will be dropped in v3.0 so this patch is no longer relevant

PR #274, opened 2024-06-20
Introduce NumberComparison validator

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| New Feature   | yes

### Description

Adds a single validator that only validates numeric input with optional lower and upper bounds.

This validator should replace `LessThan`, `GreaterThan` and `Between`

Further discussion on the reason for this validator in #272


docs/book/v2/validators/number-comparison.md
```diff
@@ -0,0 +1,75 @@
+# Number Comparison Validator
+
+`Laminas\Validator\NumberComparison` allows you to validate if a given value is a numeric value that is either
```
froschdesign:
```suggestion
`Laminas\Validator\NumberComparison` allows you to validate if a given value is a numeric value that is either:
```

docs/book/v2/validators/number-comparison.md
```diff
@@ -0,0 +1,75 @@
...
+
+- Between a min and max value
+- Greater than a min value
+- Less than a max value
+
+By default, comparisons are inclusive. 
+
+CAUTION: **Only supports number validation**
+`Laminas\Validator\NumberComparison` supports only the validation of numbers. Strings or dates can not be validated with this validator.
```
froschdesign:
```suggestion
`Laminas\Validator\NumberComparison` supports only the validation of numbers.
Strings or dates can not be validated with this validator.
```

docs/book/v2/validators/number-comparison.md
```diff
@@ -0,0 +1,75 @@
...
+
+```php
+$valid  = new Laminas\Validator\NumberComparison(['min' => 0, 'max' => 10]);
+$value  = 10;
+$result = $valid->isValid($value);
+// returns true
+```
+
+In the above example, the result is `true` due to the reason that the default search is inclusive of the border values.
+This means in our case that any value from '0' to '10' is allowed; values like '-1' and '11' will return `false`.
```
froschdesign:
```suggestion
This means in our case that any value from `0` to `10` is allowed; values like `-1` and `11` will return `false`.
```

docs/book/v2/validators/number-comparison.md
```diff
@@ -0,0 +1,75 @@
...
+    'inclusiveMin' => false,
+    'inclusiveMax' => false,
+]);
+
+$valid->isValid(10); // false
+$valid->isValid(0);  // false
+$valid->isValid(9);  // true
+```
+
+The example above is almost identical to our first example, but we now exclude the bounds as valid values; as such, the values '0' and '10' are no longer allowed and will return `false`.
```
froschdesign:
```suggestion
The example above is almost identical to our first example, but we now exclude the bounds as valid values; as such, the values `0` and `10` are no longer allowed and will return `false`.
```

docs/book/v2/validators/number-comparison.md
```diff
@@ -0,0 +1,75 @@
...
+The following options are supported for `Laminas\Validator\NumberComparison`:
+
+| Option         | Data Type | Default Value | Description                                                              |
+|----------------|-----------|---------------|--------------------------------------------------------------------------|
+| `max`          | `numeric` | `null`        | Sets the upper bound for the input.                                      |
+| `min`          | `numeric` | `null`        | Sets the lower bound for the input.                                      |
+| `inclusiveMin` | `bool`    | `true`        | Defines if the validation is inclusive of the lower bound, or exclusive. |
+| `inclusiveMax` | `bool`    | `true`        | Defines if the validation is inclusive of the upper bound, or exclusive. |
+
+## Default Behaviour
```
froschdesign:
```suggestion
## Basic Usage
```

froschdesign (changes requested):
Only minor formatting changes and a heading.

froschdesign (approved):
LGTM 👍🏻

Ocramius (approved):
:ship: thanks @gsteel!

PR #275, opened 2024-06-20
Introduce DateComparison validator

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| New Feature   | yes

### Description

Validates that a date instance or string is between, after or before given boundaries.

Supports standard string formats returned by HTML forms so that validation aligns with W3C spec.

Further ref #272, #274

### DIFF
--- a/docs/book/v2/validators/date-comparison.md
+++ b/docs/book/v2/validators/date-comparison.md
@@ -28,6 +28,9 @@ The `min` and `max` options when set must be one of the following:
 - A date string in ISO format, `YYYY-MM-DD`, i.e. '2020-01-31'
 - A date and time string in W3C format, `YYYY-MM-DDTHH:MM:SS`, i.e. '2020-01-31T12:34:56'
 
+At least one of `min`, `max` or both *must* be provided as an option or an exception will be thrown.
+It doesn't make sense to use this validator without specifying the boundaries to compare the input to.
+
 ## Default Behaviour
 
 Per default, this validator checks if a value is between `min` and `max` where both upper and lower bounds are considered valid.
--- a/docs/book/v2/validators/date-comparison.md
+++ b/docs/book/v2/validators/date-comparison.md
@@ -8,7 +8,7 @@
 
 By default, comparisons are inclusive.
 
-## Supported options
+## Supported Options
 
 The following options are supported for `Laminas\Validator\DateComparison`:
 
@@ -26,7 +26,7 @@ The `min` and `max` options when set must be one of the following:
 - A date string in ISO format, `YYYY-MM-DD`, i.e. '2020-01-31'
 - A date and time string in W3C format, `YYYY-MM-DDTHH:MM:SS`, i.e. '2020-01-31T12:34:56'
 
-## Default behaviour
+## Default Behaviour
 
 Per default, this validator checks if a value is between `min` and `max` where both upper and lower bounds are considered valid.
 
@@ -43,7 +43,7 @@ $result = $valid->isValid($value);
 In the above example, the result is `true` due to the reason that the default search is inclusive of the border values.
 This means in our case that any date between '1st January 2020' to '31st December 2020' is allowed; any other valid date will return `false`.
 
-## Min and Max behaviour
+## Min and Max Behaviour
 
 In order to validate a date that is after than a lower bound, either omit the `max` option, or set it explicitly to `null`:
 
@@ -62,7 +62,7 @@ $validator = new Laminas\Validator\DateComparison(['max' => '2020-12-31']);
 $validator->isValid('2024-06-07'); // false
 ```
 
-## Validity of Date inputs
+## Validity of Date Inputs
 
 In order to compare dates correctly, the validator converts the input to a `DateTimeInterface` object, therefore, it must be possible to parse string input as a valid date.
 
--- a/docs/book/v2/validators/date-comparison.md
+++ b/docs/book/v2/validators/date-comparison.md
@@ -12,9 +12,10 @@ By default, comparisons are inclusive.
 
 The following options are supported for `Laminas\Validator\DateComparison`:
 
-- `inclusive` _(boolean)_: Defines if the validation is inclusive of the lower and upper bounds, or exclusive. It defaults to `true`.
 - `max` _(string|DateTimeInterface)_: Sets the upper bound for the input.
 - `min` _(string|DateTimeInterface)_: Sets the lower bound for the input.
+- `inclusiveMin` _(boolean)_: Defines if the validation is inclusive of the lower bound, or exclusive. It defaults to `true`.
+- `inclusiveMax` _(boolean)_: Defines if the validation is inclusive of the upper bound, or exclusive. It defaults to `true`.
 - `inputFormat` _(string)_: Defines the expected date format if required.
 
 ## Min and Max Date Options
--- a/docs/book/v2/validators/date-comparison.md
+++ b/docs/book/v2/validators/date-comparison.md
@@ -12,11 +12,13 @@ By default, comparisons are inclusive.
 
 The following options are supported for `Laminas\Validator\DateComparison`:
 
-- `max` _(string|DateTimeInterface)_: Sets the upper bound for the input.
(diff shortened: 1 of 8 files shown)


src/DateComparison.php
```diff
@@ -0,0 +1,209 @@
...
+use function assert;
+use function get_debug_type;
+use function is_string;
+use function preg_match;
+
+/**
+ * @psalm-type OptionsArgument = array{
+ *     min?: string|DateTimeInterface|null,
+ *     max?: string|DateTimeInterface|null,
+ *     inclusive?: bool,
```
Ocramius: `inclusive` is a bit weird, considering that a `min` and `max` could be different, although it's worked around by instantiating two `DateComparison` validators

src/DateComparison.php
```diff
@@ -0,0 +1,209 @@
...
+    {
+        if (! preg_match('/^\d{4}-[0-1]\d-[0-3]\dT\d{1,2}:[0-5]\d:[0-5]\d$/', $input)) {
+            return null;
+        }
+
+        $date = DateTimeImmutable::createFromFormat('Y-m-d\TH:i:s', $input, new DateTimeZone('UTC'));
+        assert($date !== false);
+
+        return $date;
+    }
```
Ocramius, line 211: Overall OK-ish to have this, although I'd just ask the end-user to provide a `DateTimeImmutable` themselves, rather than us trying to infer one :thinking:

src/ValidatorPluginManager.php
```diff
@@ -50,6 +50,7 @@ class ValidatorPluginManager extends AbstractPluginManager
         'Csrf'                   => Csrf::class,
         'date'                   => Date::class,
         'Date'                   => Date::class,
+        'DateComparison'         => DateComparison::class,
```
Ocramius: Is it worth having these short aliases for new classes? We have `::class` since PHP 5.5, so I'd let this go away...

src/DateComparison.php
gsteel: The main reason I allowed strings - in the specific formats documented, is that I could see how you might generate a validator instance from a form input such as `<input type="date" min="2020-01-01" max="2020-12-31">`, via laminas form for example - requiring a date instance up-front would make this harder.

Ocramius: That's fine by me: it's also all tested an such, it's just a lot of complexity added, which is why I raised the question :)

src/DateComparison.php
gsteel: Do you mean it's weird because you'd want `inclusiveMin` and `inclusiveMax` options rather than just the one option for both?

Ocramius: yep, pretty much: since the validator is new, we can do it now.

gsteel: Done! Also, thanks for all the reviews today :)

Ocramius: > thanks for all the reviews today

Procrastinating work for unpaid work, yay!

Ocramius (approved):
LGTM, but would like @froschdesign's opinion first :)

src/DateComparison.php
gsteel: Same!

docs/book/v2/validators/date-comparison.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Supported Options
+
+The following options are supported for `Laminas\Validator\DateComparison`:
+
+- `max` _(string|DateTimeInterface)_: Sets the upper bound for the input.
+- `min` _(string|DateTimeInterface)_: Sets the lower bound for the input.
+- `inclusiveMin` _(boolean)_: Defines if the validation is inclusive of the lower bound, or exclusive. It defaults to `true`.
+- `inclusiveMax` _(boolean)_: Defines if the validation is inclusive of the upper bound, or exclusive. It defaults to `true`.
+- `inputFormat` _(string)_: Defines the expected date format if required.
```
froschdesign:
```suggestion
| Option         | Data Type                   | Default Value | Description                                                              |
|----------------|-----------------------------|---------------|--------------------------------------------------------------------------|
| `max`          | `string\|DateTimeInterface` | `null`        | Sets the upper bound for the input.                                      |
| `min`          | `string\|DateTimeInterface` | `null`        | Sets the lower bound for the input.                                      |
| `inclusiveMin` | `bool`                      | `true`        | Defines if the validation is inclusive of the lower bound, or exclusive. |
| `inclusiveMax` | `bool`                      | `true`        | Defines if the validation is inclusive of the upper bound, or exclusive. |
| `inputFormat`  | `string`                    | `null`        | Defines the expected date format if required.                            |
```

Output:

| Option         | Data Type                   | Default Value | Description                                                              |
|----------------|-----------------------------|---------------|--------------------------------------------------------------------------|
| `max`          | `string\|DateTimeInterface` | `null`        | Sets the upper bound for the input.                                      |
| `min`          | `string\|DateTimeInterface` | `null`        | Sets the lower bound for the input.                                      |
| `inclusiveMin` | `bool`                      | `true`        | Defines if the validation is inclusive of the lower bound, or exclusive. |
| `inclusiveMax` | `bool`                      | `true`        | Defines if the validation is inclusive of the upper bound, or exclusive. |
| `inputFormat`  | `string`                    | `null`        | Defines the expected date format if required.                            |

@gsteel 
Please recheck the default values. Thanks!

gsteel: Thanks @froschdesign - Much better!

test/DateComparisonTest.php
```diff
@@ -0,0 +1,171 @@
...
+        ]);
+    }
+
+    public function testInvalidMaxOptionIsExceptional(): void
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('Min/max date bounds must be either DateTime instances, or a string');
+        new DateComparison([
+            'max' => 'Bad news',
+        ]);
```
froschdesign, line 130: `Bad news` is a string, so the exception message is misleading.

test/DateComparisonTest.php
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    public function testThatOptionDateFormatsAreNotLenient(): void
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('Min/max date bounds must be either DateTime instances, or a string');
+        new DateComparison([
+            'max' => '2020-1-1',
+        ]);
```
froschdesign, line 139: The same here.

test/DateComparisonTest.php
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    public function testZeroBoundsIsExceptional(): void
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('At least one date boundary must be supplied');
```
froschdesign, line 145: This part is missing in the documentation. Or have I overlooked something?

froschdesign (changes requested):

test/DateComparisonTest.php
gsteel: The test is only matching the first part of the message. The message itself is very clear about the string format required.

test/DateComparisonTest.php
gsteel: Added a passage of text to the docs under `Min and Max Date Options`

test/DateComparisonTest.php
froschdesign: I had expected something like: "or a string representation of a date".

gsteel: The actual message is `Min/max date bounds must be either DateTime instances, or a string in one of the formats: "Y-m-d" for a date or "Y-m-d\TH:i:s" for date time`

gsteel: `exceptExceptionMessage` is effectively `assertStringContainsString`

froschdesign: Perfect! 👍🏻

froschdesign (approved):

gsteel:
Thank you for the reviews @froschdesign and @Ocramius 👍

PR #278, opened 2024-06-21
Refactor `StringLength` validator

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes
| QA            | yes

### Description

- remove all option setters and getters
- constructor options now only accepts a documented array shape
- Provides better handling of malformed utf-8 and other string handling issues in underlying string wrappers
- adds types to parameters, properties and methods

### DIFF
--- a/test/StringLengthTest.php
+++ b/test/StringLengthTest.php
@@ -5,174 +5,92 @@ declare(strict_types=1);
 namespace LaminasTest\Validator;
 
 use Laminas\Validator\Exception\InvalidArgumentException;
+use Laminas\Validator\Exception\RuntimeException;
 use Laminas\Validator\StringLength;
 use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\TestCase;
 
-use function array_keys;
-use function ini_set;
+use function chr;
+use function restore_error_handler;
+use function set_error_handler;
+
+use const E_WARNING;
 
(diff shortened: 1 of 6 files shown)


test/StringLengthTest.php
```diff
@@ -5,174 +5,92 @@
...
-            'valid; minimum: 2; maximum: 3; abc' => [[2, 3], true, 'abc'],
-
-            'invalid; minimum: 2; maximum: 3; a'    => [[2, 3], false, 'a'],
-            'invalid; minimum: 2; maximum: 3; abcd' => [[2, 3], false, 'abcd'],
-
-            'valid; minimum: 3; maximum: 3; äöü'       => [[3, 3],    true, 'äöü'],
-            'valid; minimum: 6; maximum: 6; Müller'    => [[6, 6],    true, 'Müller'],
-            'valid; minimum: null; maximum: 6; Müller' => [[null, 6], true,'Müller'],
-            // phpcs:enable
+            [0, null, 'utf-8', '', true, null],
```
boesing: I actually liked the named data providers as these are listed when tests are failing. makes finding the failing data provider easier as one can search for the name rather than checking all arguments.

boesing (approved):

test/StringLengthTest.php
gsteel: Agreed - this is a bit lazy of me 😬

PR #279, opened 2024-06-21
UUID Validator Cleanup

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| QA            | yes

### Description

- Add types
- Use `ramsey/uuid` to generate a range of random UUIDs for tests
- regex constant is now private

### DIFF
--- a/test/UuidTest.php
+++ b/test/UuidTest.php
@@ -4,14 +4,10 @@ declare(strict_types=1);
 
 namespace LaminasTest\Validator;
 
-use Generator;
 use Laminas\Validator\Uuid;
 use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\TestCase;
-use Ramsey\Uuid\Uuid as UuidGen;
 use stdClass;
-
-use function sprintf;
 
 final class UuidTest extends TestCase
 {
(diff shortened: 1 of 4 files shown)


test/UuidTest.php
```diff
@@ -46,23 +47,25 @@ public function testInvalidUuid(mixed $uuid, string $expectedMessageKey): void
...
-            'version-2' => ['ff6f8cb0-c57d-21e1-9b21-0800200c9a66'],
-            'version-3' => ['ff6f8cb0-c57d-31e1-9b21-0800200c9a66'],
-            'version-4' => ['ff6f8cb0-c57d-41e1-9b21-0800200c9a66'],
-            'version-5' => ['ff6f8cb0-c57d-51e1-9b21-0800200c9a66'],
-            'uppercase' => ['FF6F8CB0-C57D-11E1-9B21-0800200C9A66'],
-        ];
+        yield 'zero-fill' => ['00000000-0000-0000-0000-000000000000'];
+
+        for ($i = 0; $i < 10; $i++) {
+            yield sprintf('v1-#%d', $i) => [UuidGen::uuid1()->toString()];
```
boesing: hm, what was the problem with the static UUIDs? I mean, these were valid?
I don't mind the additional dev-dependency but that makes tests less reproducible due to the randomly generated UUIDs.

boesing (approved):

test/UuidTest.php
gsteel: It just didn't make sense to me to keep validating static UUIDs - those values are never gonna show up in the wild and the lib should validate _any_ UUID thown at it. I figured that this data set would make for more robust tests 🤷‍♂️

boesing: this actually tests ramseys implementation. so if there is a bug, that might end up reported in PRs in this repository.
Thats why I personally try to avoid tests like this as "unit" tests.

I mean, you could create integration tests in a dedicated space so that these are detached but I'd keep it simple.
UUIDs are just strings, no matter if these are out there in the wild or not. If the string is valid, it makes no diff.
What makes a diff is having random uuids in test results, but thats already reported in my initial comment.

However, I do not mind this, but I personally would not implement a unit test like this.

gsteel: Thanks - I'll take your lead on this and revert the extra dependency 👍

PR #280, opened 2024-06-21
Refactor `Regex` validator

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | arguably
| BC Break      | yes
| QA            | yes

### Description

- Fails validation for non strings
- Removes getters and setters
- Adds types
- Validates the pattern in the constructor instead of during validation meaning we can drop one of the validation error constants

Closes #179

### DIFF
--- a/src/Regex.php
+++ b/src/Regex.php
@@ -4,32 +4,29 @@ declare(strict_types=1);
 
 namespace Laminas\Validator;
 
-use Laminas\Stdlib\ArrayUtils;
-use Laminas\Stdlib\ErrorHandler;
-use Traversable;
+use Laminas\Validator\Exception\InvalidArgumentException;
 
-use function array_key_exists;
-use function is_array;
-use function is_float;
-use function is_int;
 use function is_string;
 use function preg_match;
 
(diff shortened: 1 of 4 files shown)


src/Regex.php
```diff
@@ -38,98 +35,48 @@ final class Regex extends AbstractValidator
...
-        $status = preg_match($this->pattern, (string) $value);
-        ErrorHandler::stop();
-        if (false === $status) {
-            $this->error(self::ERROROUS);
-            return false;
-        }
+        $status = preg_match($this->pattern, $value);
 
-        if (! $status) {
+        if ((bool) $status === false) {
```
boesing, line 79: Feels kinda weird that there is a type-cast, even tho I use PHP on a daily basis, I never trust the way it parses the code.
Just by looking at it, I do not know if it applies the type-cast to the result of `$status === false` (which is already bool) or to `status` prior explicit check. Though I'd try to avoid this kind of type-cast but thats just my inner monk.

boesing (approved):

PR #286, opened 2024-06-23
Refactor `Explode` validator

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes
| QA            | yes

### Description

- Remove option getters and setters
- Only accept an array for configuration
- Add types everywhere
- Changes handling of error messages
- fail validation for non-string arguments
- Allow the composed validator option to be a class-string or alias

Validator error messages should be `array<string, string>`. This validator would yield `list<array<string, string>>` on validation failure. The validation error was also largely pointless because repeated errors would be squashed internally by AbstractValidator leaving only one element in the list anyhow.

This validator now yields a new error message, something like `"5 items were invalid: composed validator error messages"`

### DIFF
--- a/docs/book/v3/validators/explode.md
+++ b/docs/book/v3/validators/explode.md
@@ -39,7 +39,7 @@ $explodeValidator = new Laminas\Validator\Explode([
     'validator' => [
         'name' => Laminas\Validator\InArray::class,
         'options' => [
-            'haystack' => 'a;b;c'
+            'haystack' => ['a', 'b', 'c']
         ],
     ],
     'valueDelimiter' => ';',
--- a/docs/book/v3/validators/explode.md
+++ b/docs/book/v3/validators/explode.md
@@ -8,9 +8,10 @@ array.
 The following options are supported for `Laminas\Validator\Explode`:
 
 - `valueDelimiter`: Defines the delimiter used to explode values from an array.
-  It defaults to `,`. If the given value is an array, this option isn't used.
+  It defaults to `,`.
 - `validator`: Sets the validator that will be executed on each exploded item.
-  This may be a validator instance, or a validator service name.
+  This may be a validator instance, a validator service name, or a "specification" array.
+- `validatorPluginManager`: The validator plugin manager in use in your application. If this plugin manager is not provided, a plugin manager will be created with the default configuration.
 
 ## Basic usage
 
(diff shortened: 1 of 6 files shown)


src/Explode.php
```diff
@@ -5,82 +5,64 @@
...
+        parent::__construct($options);
     }
 
-    /**
-     * Set validator plugin manager
-     *
-     * @return void
-     */
-    public function setValidatorPluginManager(ValidatorPluginManager $pluginManager)
+    private function pluginManager(): ValidatorPluginManager
```
froschdesign: Did I miss something, otherwise I would have expected `getPluginManager` here?!

gsteel: This patch drops `PluginManagerAwareInterface`, requiring the user to supply a plugin manager to the constructor, if they want to use a specific one.

The added `private` method `pluginManager()` lazy loads a new plugin manager instance if the user did not supply one to the constructor. The diff is probably a bit of a mess - a lot has been removed.

froschdesign: The function is clear, but I mean the name of the method itself.

gsteel: Ah, OK - I guess this is personal habit - `pluginManager()` is more command-like IMO - the name change is unnecessary so I'll revert.

froschdesign: The name is okay but I was only thinking about the consistency. So far it was always with `get`.

psalm-baseline.xml
```diff
@@ -540,28 +539,17 @@
...
+    <LessSpecificReturnStatement>
+      <code><![CDATA[$this->getPluginManager()->build($name, $options)]]></code>
+    </LessSpecificReturnStatement>
+    <MixedInferredReturnType>
+      <code><![CDATA[ValidatorInterface]]></code>
+    </MixedInferredReturnType>
+    <MixedReturnStatement>
+      <code><![CDATA[$this->getPluginManager()->build($name, $options)]]></code>
+      <code><![CDATA[$this->getPluginManager()->get($validator)]]></code>
+      <code><![CDATA[$this->getPluginManager()->get($validator)]]></code>
```
Ocramius, line 551: Is `getPluginManager()` not appropriately typed here? :thinking:

docs/book/v3/validators/explode.md
```diff
@@ -25,24 +26,25 @@ $explodeValidator = new Laminas\Validator\Explode([
...
-    'haystack' => ['joseph@test.com', 'mark@test.com', 'lucia@test.com'],
-]);
-
 $explodeValidator = new Laminas\Validator\Explode([
-    'validator' => $inEmailListValidator,
-    'valueDelimiter' => ','
+    'validator' => [
+        'name' => Laminas\Validator\InArray::class,
+        'options' => [
+            'haystack' => 'a;b;c'
```
Ocramius: This example seems wrong: `a;b;c` should be an `array` `['a', 'b', 'c']` passed to `InArray`?

test/ExplodeTest.php
```diff
@@ -4,201 +4,152 @@
...
-            ['foo', ',', true, 1, false, ['X'], false],
-            [['a', 'b'], null, false, 2, true, [], true],
-            [['a', 'b'], null, false, 2, false, ['X'], false],
-            ['foo', null, false, 1, true, [], true],
-            [1, ',', false, 1, true, [], true],
-            [null, ',', false, 1, true, [], true],
-            [new stdClass(), ',', false, 1, true, [], true],
-            [new ArrayObject(['a', 'b']), null, false, 2, true, [], true],
-        ];
-    }
```
Ocramius, line 66: Unsure why this stuff is gone: was it just wrong?

Ocramius (changes requested):
TIL: there is an `Explode` validator :D

psalm-baseline.xml
gsteel: These should go away once we have SMv4 support merged

test/ExplodeTest.php
gsteel: This data provider was for a mostly messy test with too much mocking. All the covered cases were dealt with in specific standalone tests.

Ocramius (approved):
Thanks @gsteel!

PR #288, opened 2024-06-24
Refactor `Timezone` validator

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | yes
| BC Break      | yes
| QA            | yes

### Description

- Removes all option setters and getters
- Allow only an array for the options passed to the constructor
- Only accept an `int-mask` for the `type` option
- Add types
- Make TZ abbreviation comparison case-insensitive


Ocramius (approved):

Ocramius:
Baseline needs adjusting, I think

Ocramius:
Thanks, @gsteel!

PR #291, opened 2024-06-24
Refactor `Barcode` validator

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes
| QA            | yes

### Description

- Removes all options setters and getters
- Allows class-string for the adapter
- Drops options for the adapters because none of them accepted options
- Drop the abstract adapter and simplify adapter interface
- Remove most getters/setters for adapters
- extract useful checksum methods from the abstract adapter to an internal utility class


src/Barcode/Code39ext.php
```diff
@@ -4,15 +4,25 @@
...
+    }
+
+    public function hasValidCharacters(string $value): bool
+    {
+        return Util::isAscii128($value);
+    }
+
+    public function hasValidChecksum(string $value): bool
+    {
+        return true;
```
froschdesign, line 21: How can this be `true` when the tests say the check is missing? 🤔

```php
// @TODO: CODE39 EXTENDED CHECKSUM VALIDATION MISSING
``` 
https://github.com/laminas/laminas-validator/pull/291/files#diff-82f4f4e291d22c9cdc2edf53bcb08cf222fec6369e2b1e42f1bdb15868f6a44bR194

gsteel: So there are at least 2 adapters that don't have any checksum validation and a TODO like this.

In 2.x these adapters would have passed validation regardless of whether cksum validation was on or off, so returning `true` in these cases preserves existing behaviour, regardless of how pointless these adapters are.

I'm not going to investigate how to implement the cksum for these adapters, so should we leave them as-is, or deprecate in 2.x and remove them here?

froschdesign: > …so should we leave them as-is…

This and we can create issue reports for these to-dos.

froschdesign (approved):
LGTM 👍🏻

PR #297, opened 2024-06-28
Add parameter and return types to `ValidatorInterface::isValid()`

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| QA            | yes

### Description

The main change here is changing the signature of `ValidatorInterface::isValid()` to `public function isValid(mixed $value): bool;`

The patch includes some other miscellaneous clean up:

- Add property types to common props such as `$messageTemplates`
- converts some protected static props to private constants


froschdesign (approved):
LGTM 👍🏻

PR #298, opened 2024-06-30
Refactor `Hostname` validator

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes
| QA            | yes

### Description

- Only allow a well known array shape as options for the constructor
- Add types everywhere
- Replace internal options array with RO props
- Remove option setters and getters

### DIFF
--- a/docs/book/v3/validators/hostname.md
+++ b/docs/book/v3/validators/hostname.md
@@ -13,8 +13,7 @@ The following options are supported for `Laminas\Validator\Hostname`:
   [See below](#validating-different-types-of-hostnames) for details.
 - `idn`: Defines if IDN domains are allowed or not. This option defaults to
   `true`.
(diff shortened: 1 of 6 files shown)


docs/book/v3/validators/hostname.md
```diff
@@ -63,7 +62,7 @@ To check for IP addresses only, you can use the example below:
 ```php
 use Laminas\Validator\Hostname;
 
-$validator = new Hostname(Hostname::ALLOW_IP);
+$validator = new Hostname(['allow' => Hostname::ALLOW_IP]);
```
Ocramius, line 65: Is the single parameter approach still working?

Ocramius (approved):

docs/book/v3/validators/hostname.md
gsteel: Nope, only an options array is accepted

PR #303, opened 2024-07-02
Change Malformed Multi-Byte input handling in `StringLength`

### LINKED ISSUE #296: [RFC]: StringLength validator handling of malformed multibyte input
## RFC

| Q                   | A      |
| ------------------- | ------ |
| Proposed Version(s) | 3.0.0  |
| BC Break?           | No |

### Background

In #278 the string length validator is changed so that malformed multibyte input, rejected by the underlying string wrapper, will cause a predictable exception.

Previously, in the 2.x series, errors, warnings and exceptions would propagate directly from function calls such as `mb_strlen` and the string wrapper.

### Questions

Before releasing 3.0, it's worth addressing whether it would be a better course of action for this type of issue to become a validation failure or not

If consesus is that this condition should be a validation failure instead of an exception, should errors and warnings from iconv/mbstring/intl be silenced?

gsteel:
Instead of throwing an exception for malformed UTF-8, trigger a validation failure

Closes #296

### DIFF
--- a/src/StringLength.php
+++ b/src/StringLength.php
@@ -6,7 +6,6 @@ namespace Laminas\Validator;
 
 use Laminas\Stdlib\StringUtils;
 use Laminas\Validator\Exception\InvalidArgumentException;
-use Laminas\Validator\Exception\RuntimeException;
 use Throwable;
 
 use function is_string;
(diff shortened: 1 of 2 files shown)


froschdesign (approved):
LGTM 👍🏻

PR #316, opened 2024-07-03
Refactor `ISBN` validator

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes
| QA            | yes

### Description

- Drops the `separator` option in favour of stripping them from the input
- Inlines checksum calculation, removing unnecessary classes
- Adds types
- Accept only an options array to the constructor
- Simplify tests with data providers


froschdesign (approved):
Nice improvement! 👍🏻

PR #320, opened 2024-07-03
Refactor `Identical` validator

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes
| QA            | yes

### Description

- Removes option getters and setters
- Accept only an option array to the constructor
- Add types

### DIFF
--- a/src/Identical.php
+++ b/src/Identical.php
@@ -4,19 +4,19 @@ declare(strict_types=1);
 
 namespace Laminas\Validator;
 
-use ArrayAccess;
-use Laminas\Stdlib\ArrayUtils;
-use Traversable;
-
-use function array_key_exists;
-use function get_debug_type;
 use function is_array;
 use function is_int;
 use function is_string;
 use function key;
-use function sprintf;
 use function var_export;
 
+/**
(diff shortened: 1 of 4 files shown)


src/Identical.php
```diff
@@ -33,172 +33,78 @@ final class Identical extends AbstractValidator
...
-    /**
-     * Returns the literal parameter
-     *
-     * @return bool
-     */
-    public function getLiteral()
-    {
-        return $this->literal;
-    }
+        $this->tokenString = is_array($token) ? var_export($token, true) : (string) $token;
```
froschdesign, line 55: Did I miss something? Where is this used?
Shouldn't this be in a message?

gsteel: It's available for use in message placeholders, but it's not used here. Shall I kill it?

froschdesign: Leave it there, because the class is marked as final and writing a new class only for the reason to use the token in a message seems too much to me.

froschdesign (approved):
LGTM 👍🏻

PR #324, opened 2024-07-03
Refactor `EmailAddress` Validator

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes
| QA            | yes

### Description

- Remove all setters and getters
- Add types
- Update docs

PR #325, opened 2024-07-03
Require `ext-intl`

gsteel:
### Description

The email address validator relies on `ext-intl` for important parts of its job.

Also, the script that updates TLDs for the Hostname validator needs `idn_to_utf8`

This patch requires it and removes the `extension_loaded` calls

Cut from #324


froschdesign (approved):
LGTM 👍🏻

PR #326, opened 2024-07-03
`EmailAddress` validator deprecations

gsteel:
Deprecates all option setters and getters

See #324


froschdesign (approved):
If the method `setMessage` also remains in the new version, then: 👍🏻

gsteel:
> If the method `setMessage` also remains in the new version, then: 👍🏻

Once I've refactored all the remaining validators, I'll work on `AbstractValidator`, at which point this might change, but I'll deal with that when I get there.

PR #329, opened 2024-07-04
Fix TLD updater script

gsteel:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes
| QA            | yes

### Description

In #298 the `$validTlds` property declaration was changed but the regex in the script was not.

This patch also removes the check for `ext-intl` which is already available in CI, and instead outputs an error when not available. This has the side effect of 100% type coverage.

Also relevant: #325

### DIFF
--- a/bin/update_hostname_validator.php
+++ b/bin/update_hostname_validator.php
@@ -2,8 +2,6 @@
 
 declare(strict_types=1);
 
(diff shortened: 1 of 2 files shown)


froschdesign (approved):
LGTM 👍🏻

PR #330, opened 2024-07-04
Refactor `Date` and `DateStep` Validators

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes
| QA            | yes

### Description

- Remove setters and getters
- Add types
- Only accepts options arrays to constructors
- Removes unused timezone option from DateStep
- Clean up and QA
- Changes default format in DateStep from `DATE_ISO8601` (Deprecated) to `DATE_W3C`

### DIFF
--- a/src/DateStep.php
+++ b/src/DateStep.php
@@ -46,7 +46,7 @@ final class DateStep extends Date
     /**
      * Default format constant
      */
-    public const FORMAT_DEFAULT = DateTime::ISO8601;
+    public const FORMAT_DEFAULT = DateTimeInterface::ATOM;
 
     /**
      * Validation failure message template definitions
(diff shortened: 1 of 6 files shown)


src/DateStep.php
```diff
@@ -46,7 +46,7 @@ final class DateStep extends Date
     /**
      * Default format constant
      */
-    public const FORMAT_DEFAULT = DateTime::ISO8601;
+    public const FORMAT_DEFAULT = DateTimeInterface::W3C;
```
froschdesign: The PHP documentation and the deprecation message say `ATOM`:

> This format is not compatible with ISO-8601, but is left this way for backward compatibility reasons. Use DateTime::ATOM or DATE_ATOM for compatibility with ISO-8601 instead.

But W3C is the same:

```
public const ATOM = 'Y-m-d\TH:i:sP';
public const W3C  = 'Y-m-d\TH:i:sP';
```

gsteel: 👍 I'll switch to `ATOM`

froschdesign (approved):
LGTM 👍🏻

PR #338, opened 2024-07-04
Remove unused exception classes

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| QA            | yes

### Description

Also changes exception inheritance so that exceptions can all be final


froschdesign (approved):
LGTM 👍🏻

PR #340, opened 2024-07-05
Refactor `MimeType` validator and descendants

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes
| QA            | yes

### Description

- Requires `ext-fileinfo` - The extension is enabled by default, so it's reasonable to assume that most installations will have it.
- Requires `laminas-diactoros` in `dev` so that we can more easily test with `UploadedFileInterface`
- Introduce new `@internal` class `FileInformation` which, eventually, all validators in the `\File` namespace will use. This class represents a file that exists and extracts actual mime detection from the validators.
- Removes `Magic File` options and handling completely. `finfo` does not need a custom magic file - it's safer and easier to use the magic file the PHP uses internally.
- Removes the `disableMagicFile` option for obvious reasons
- Removes the `enableHeaderCheck` option which looks like it might have been some legacy hangover that at some point was looking at request headers perhaps.
- Allows only structured arrays as validator constructor arguments
- Requires the `mimeType` option for the `MimeType` validator and the `ExcludeMimeType` validator
- Drops all option setters and getters - options can now only be set during construction
- Drops compatibility with the legacy `Laminas\File\Transfer` api

I can't currently get mime type detection to fail. No matter what kind of corrupted files or weird filename extensions, mime detection always returns a string.


froschdesign (approved):
Many changes were made, so it took a little longer.
Same quality as before, so: 👍🏻

gsteel:
Thanks @froschdesign

PR #341, opened 2024-07-05
Clean up usage of `$this->options`

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

`StringLength::$options` was completely unused

`Ip` and `NumberComparison` validators have been refactored internally to use typed instance properties instead of `$options` array


froschdesign (approved):
LGTM 👍🏻

PR #343, opened 2024-07-05
Deprecate setters and getters in File based validators

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes

### Description

Deprecates option setters and getters in File based validators


froschdesign (approved):
It's great that so much can be removed. 👍🏻

PR #345, opened 2024-07-05
 Refactor `WordCount` validator

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes

### Description

 - Exceptions no longer thrown for non-existent or un-readable files (Validation failure instead)
 - Constructor only accepts an options array
 - Drops legacy `Laminas\File\Transfer` compat


froschdesign (approved):
LGTM 👍🏻

PR #348, opened 2024-07-05
Refactor `Size` and `FilesSize` Validators

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes
| QA            | yes

### Description

- Removes options setters and getters
- Adds types
- Drops compat with legacy `Laminas\File\Transfer` api
- Moves SI unit parsing and formatting to FileInformation class

### DIFF
--- a/src/File/FileInformation.php
+++ b/src/File/FileInformation.php
@@ -10,8 +10,10 @@ use Psr\Http\Message\UploadedFileInterface;
 use function assert;
 use function basename;
 use function file_exists;
+use function filesize;
 use function finfo_open;
 use function is_array;
+use function is_int;
 use function is_readable;
 use function is_string;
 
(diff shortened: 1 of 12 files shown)


src/File/FileInformation.php
```diff
@@ -109,4 +136,77 @@ public static function isPossibleFile(mixed $value): bool
...
+
+        assert(is_string($suffix));
+
+        return round($size, 2) . $suffix;
+    }
+
+    /**
+     * Convert an SI unit to bytes
+     */
+    public static function siUnitToBytes(string $size): int
```
Ocramius: Probably worth extracting to its own value object (marked `@internal` in this library), so it can be tested without involving `FileInformation`.

While you did already test this, I don't think it should be mixed in: it's a ton of code.

gsteel: Good call. All done :)

Ocramius (approved):

PR #351, opened 2024-07-09
Refactor `Exists` and `NotExists` validators

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes
| QA            | yes

### Description

- Removes all setters and getters
- Removes inheritance so both classes can be final
- Removes compat with legacy `Laminas\File\Transfer` api
- Improves docs and tests

### DIFF
--- a/docs/book/v3/validators/file/exists.md
+++ b/docs/book/v3/validators/file/exists.md
@@ -9,7 +9,8 @@ This validator is inversely related to the [NotExists validator](not-exists.md).
 
 The following set of options are supported:
 
-- `directory`: Array of directories, or comma-delimited string  of directories.
+- `directory`: Array of directories, or comma-delimited string of directories.
+- `all`: A boolean that when `true` _(default)_ requires that the filename is present in **all** the listed directories 
 
 ## Usage Examples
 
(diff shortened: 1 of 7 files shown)


docs/book/v3/validators/file/exists.md
```diff
@@ -9,27 +9,64 @@ This validator is inversely related to the [NotExists validator](not-exists.md).
...
+- `all`: A boolean that when `true` _(default)_ requires that the filename is present in **all** the listed directories 
 
 ## Usage Examples
 
 ```php
 use Laminas\Validator\File\Exists;
 
 // Only allow files that exist in ~both~ directories
-$validator = new Exists('/tmp,/var/tmp');
+$validator = new Exists(['directory' => '/tmp,/var/tmp']);
```
froschdesign, line 21:
```suggestion
$validator = new Laminas\Validator\File\Exists(['directory' => '/tmp,/var/tmp']);
```

src/File/Exists.php
```diff
@@ -5,180 +5,123 @@
...
+    public function __construct(array $options = [])
     {
-        if (is_string($options)) {
-            $options = explode(',', $options);
-        }
-
-        if (is_array($options) && ! array_key_exists('directory', $options)) {
-            $options = ['directory' => $options];
-        }
+        $this->directories         = $this->resolveDirectories($options['directory'] ?? null);
```
froschdesign, line 59: It is a little confusing that the `directory` option can also be used for multiple directories. And internally it is called `directories`. 🤔

gsteel: The plural is more correct considering the option allows multiple directories, but the key itself I kept as `directory` for BC. I could change the option key but it's more to document and more for users to deal with. Also, I'd rather not change it add extra checks for legacy keys because it's more code and noise in the constructor.

What do you think is the best approach?

froschdesign: My first thought was to allow both options, but this doesn't make it any better:

```php
$validator = new Exists([
    'directory' => '/tmp',
    'directories' => ['/var/tmp', '/var/www/tmp'],
    'all' => false,
]);
```

> …but the key itself I kept as `directory` for BC.

So I agree with this argument.

gsteel:
Happy for this one to be merged @froschdesign ?

froschdesign (approved):
LGTM 👍🏻

PR #353, opened 2024-07-10
Deprecate Unnecessary File Hash Validators

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes

### Description

The `Hash` validator is sufficient for comparing the hash of a given file.

`Crc32`, `Md5` and `Sha1` validators can easily be replaced with the parent `Hash` validator and the correct options.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -2466,6 +2466,24 @@
       <code><![CDATA[Db\RecordExists::class]]></code>
       <code><![CDATA[Db\RecordExists::class]]></code>
       <code><![CDATA[Db\RecordExists::class]]></code>
+      <code><![CDATA[File\Crc32::class]]></code>
+      <code><![CDATA[File\Crc32::class]]></code>
+      <code><![CDATA[File\Crc32::class]]></code>
+      <code><![CDATA[File\Crc32::class]]></code>
+      <code><![CDATA[File\Crc32::class]]></code>
+      <code><![CDATA[File\Crc32::class]]></code>
+      <code><![CDATA[File\Md5::class]]></code>
+      <code><![CDATA[File\Md5::class]]></code>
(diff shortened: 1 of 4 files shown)


froschdesign (approved):
Go for it! 👍🏻

InvisibleSmiley:
I found that in ^2.61 using `Hash` instead of e.g. `Sha1` only works if you specify the `algorithm` option *first*:

```php
// works
new Hash(['algorithm' => 'sha1', 'hash' => 'foo']);
// fails
new Hash(['hash' => 'foo', 'algorithm' => 'sha1']);
```

This is because the parent constructor calls setters in the order that the keys appear in the options array.

I'd assume that ^3.0 doesn't have that issue but we cannot use this yet because our project also requires laminas/laminas-inputfilter.

gsteel:
@InvisibleSmiley - Can you create a new issue for this? Is it a regression, or a bug that has never previously been found? If you have time for a failing test case and associated fix, that would be much appreciated 👍

Work is in progress on `laminas-filter:3.0`, once that is released, we'll be able to start work on `input-filter` and then eventually `form`…

InvisibleSmiley:
> Can you create a new issue for this? 

Filed #398.

> Is it a regression, or a bug that has never previously been found?

Probably not found before; exists since at least 2.15.0 (minimum version I could install due to my project dependencies).

So certainly not a recent regression. I guess people used the dedicated algorithm classes until now so this went unnoticed. Assuming that 3.x is OK (no setters, no worries?), this will only be an issue for people trying to migrate in advance.

> If you have time for a failing test case and associated fix, that would be much appreciated 👍

Well my first question would be whether you intend to ship a 2.x fix. Otherwise there would be no point in working on one. ;-) But we can discuss that over there.

PR #354, opened 2024-07-10
Refactor `File\Hash` validator

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes
| QA            | yes

### Description

- Drop compat with legacy `Laminas\File\Transfer` api
- Add types
- Remove all option getters and setters
- Only accept an options array to the constructor
- Remove the ability to validate multiple algorithms in the same instance. This feature was not tested and also nonsensical. To validate the `sha1` and the `md5` of a file, chain 2 validator instances.
- Remove the deprecated `Md5`, `Sha1` and `Crc32` file validators


src/File/Hash.php
```diff
@@ -43,129 +35,70 @@ class Hash extends AbstractValidator
...
-     * @throws Exception\InvalidArgumentException
-     * @return $this Provides a fluent interface
-     */
-    public function addHash($options)
-    {
-        if (is_string($options)) {
-            $options = [$options];
-        } elseif (! is_array($options)) {
-            throw new Exception\InvalidArgumentException('False parameter given');
+        $algorithm = strtolower($options['algorithm'] ?? 'crc32');
```
froschdesign, line 60: PHP itself works without constants for the algorithms, so it should also be sufficient here.

froschdesign (approved):
Much cleaner! 👍🏻

PR #355, opened 2024-07-11
Refactor `Extension` and `ExcludeExtension` Validators

### LINKED ISSUE #98: File extension comparison treats some distinct extensions as equal

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes
| QA            | yes

### Description

- Remove compat with legacy `Laminas\File\Transfer` api
- Remove all setters and getters
- Adjust inheritance so both classes can be final
- Only accept an options array to the constructor
- Add new error message for completely invalid values (non-string, non-upload) instead of throwing an exception


src/File/Extension.php
```diff
@@ -4,240 +4,204 @@
...
      */
-    public function __construct($options = null)
+    public function __construct(array $options)
     {
-        if ($options instanceof Traversable) {
-            $options = ArrayUtils::iteratorToArray($options);
-        }
+        $this->caseSensitive        = $options['case'] ?? false;
+        $this->allowNonExistentFile = $options['allowNonExistentFile'] ?? false;
+        $this->extensions           = self::resolveExtensionList($options['extension'] ?? []);
```
froschdesign, line 68: The same as before: the singular is okay.

froschdesign (approved):
LGTM 👍🏻

gsteel:
Thank you @froschdesign 🙏

Issue #347, opened 2024-07-05
[RFC]: Deprecate and remove the `Upload` validator


gsteel:
## RFC

| Q                   | A      |
| ------------------- | ------ |
| Proposed Version(s) | 3.0.0  |
| BC Break?           | Yes |

### Goal

Remove seeimingly unnecessary validator

### Background

There are 2 validators for validating successful upload:

- https://github.com/laminas/laminas-validator/blob/2.61.x/src/File/Upload.php
- https://github.com/laminas/laminas-validator/blob/2.61.x/src/File/UploadFile.php

`UploadFile` accepts a value to `isValid` representing a single file and returns true when the file appears to be successfully uploaded.

`Upload` requires a list of all uploaded files to its constructor, and subsequent calls to `isValid` represent the form element name/key it seems. i.e. `isValid('myfile')`

It seems odd that `Upload` needs the input as an argument to its constructor (Or option) rather than to the `isValid` method.

### Considerations

I might be missing the point of the `Upload` validator. It looks like something from the days of `Laminas\File\Transfer` and `Laminas\Http`.

If anyone can shed any light on why it works the way it does, I'm all ears.

### Proposal(s)

Deprecate the validator in 2.61.0 and remove it in 3.0.0

cc @laminas/technical-steering-committee 

Edit: The `Upload` validator works directly with the `$_FILES` super-global 🤢

froschdesign:
@weierophinney 
> ### Considerations
> I might be missing the point of the `Upload` validator. It looks like something from the days of `Laminas\File\Transfer` and `Laminas\Http`.
> 
> If anyone can shed any light on why it works the way it does, I'm all ears.

Do you have a comment on this?

gsteel:
Closed via #362

PR #358, opened 2024-07-11
Refactor `AbstractValidator`

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes
| QA            | yes

### Description

Removes methods from `TranslatorAwareInterface` and the implementations in `AbstractValidator`:

- `hasTranslator`
- `setTranslatorEnabled`
- `isTranslatorEnabled`
- `setTranslatorTextDomain`
- `getTranslatorTextDomain`

Leaving only `setTranslator` and `getTranslator`

Enabling and disabling translation can be done via options. The getters are unnecessary.

Removes methods from `AbstractValidator`:

- `getOption`
- `getOptions`
- `setOptions`
- `getMessageVariables`
- `getMessageTemplates`
- `setMessages`
- `setValueObscured`
- `isValueObscured`
- `getDefaultTranslator`
- `hasDefaultTranslator`
- `getDefaultTranslatorTextDomain`
- `getMessageLength`

Reduces the API surface of `AbstractValidator` making a number of previously `protected` methods `private`.

Removes `abstractOptions` property in favour of protected or private instance properties.

Remove a number of useless tests that verified the content of internal properties, i.e. ensuring that `messageTemplates` were identical to the declaration.

Appends the options relevant to the abstract class such as `translator`, `messages` etc to all the options declarations in concrete validators. Psalm does not support unions of array shapes in type declarations.

Adds more types: Down to ~230 psalm issues and 99.2% type coverage

### DIFF
--- a/src/AbstractValidator.php
+++ b/src/AbstractValidator.php
@@ -89,13 +89,13 @@ abstract class AbstractValidator implements
     private bool $valueObscured = false;
 
     /** Whether translation should be enabled or not */
-    private bool $translatorEnabled;
+    private bool $translatorEnabled = true;
 
     /** The text domain for translations */
-    private string $translatorTextDomain;
+    private string $translatorTextDomain = 'default';
 
     /** A custom translator, the default translator, or null */
-    private TranslatorInterface|null $translator;
+    private TranslatorInterface|null $translator = null;
 
     /**
      * Error messages that have occurred during the last validation
--- a/src/AbstractValidator.php
+++ b/src/AbstractValidator.php
@@ -4,42 +4,39 @@ declare(strict_types=1);
 
 namespace Laminas\Validator;
 
-use Laminas\Stdlib\ArrayUtils;
 use Laminas\Translator\TranslatorInterface;
-use Traversable;
+use Laminas\Validator\Exception\InvalidArgumentException;
 
 use function array_key_exists;
 use function array_keys;
 use function array_unique;
-use function current;
+use function assert;
 use function implode;
 use function is_array;
+use function is_bool;
 use function is_object;
 use function is_string;
 use function key;
 use function method_exists;
+use function property_exists;
+use function sprintf;
 use function str_repeat;
 use function str_replace;
 use function strlen;
 use function substr;
-use function ucfirst;
 use function var_export;
 
 use const SORT_REGULAR;
 
 /**
  * @psalm-type AbstractOptions = array{
- *     messages: array<string, string>,
- *     messageTemplates: array<string, string>,
- *     messageVariables: array<string, mixed>,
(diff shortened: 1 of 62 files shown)


gsteel:
Need to look at docs - also tests are failing because of the `Upload` validator: See #347

src/AbstractValidator.php
```diff
@@ -55,320 +52,217 @@ abstract class AbstractValidator implements
...
+    public function __construct(array $options = [])
     {
-        /** @psalm-suppress DocblockTypeContradiction */
-        if (! is_array($options) && ! $options instanceof Traversable) {
-            throw new Exception\InvalidArgumentException(__METHOD__ . ' expects an array or Traversable');
-        }
+        $valueObscured        = $options['valueObscured'] ?? false;
+        $translatorEnabled    = $options['translatorEnabled'] ?? true;
+        $translatorTextDomain = $options['translatorTextDomain'] ?? self::$defaultTranslatorTextDomain;
+        $translator           = $options['translator'] ?? self::$defaultTranslator;
```
Ocramius, line 122: should we keep this static usage? :thinking:

src/DateComparison.php
```diff
@@ -21,6 +22,11 @@
  *     inclusiveMin?: bool,
  *     inclusiveMax?: bool,
  *     inputFormat?: string|null,
+ *     messages?: array<string, string>,
+ *     translator?: TranslatorInterface|null,
+ *     translatorTextDomain?: string|null,
+ *     translatorEnabled?: bool,
+ *     valueObscured?: bool,
```
Ocramius, line 29: This type is repeated for every validator: is this something we can prevent somehow, or does `&AbstractOptions` not work here?

Ocramius (approved):
Changes make sense to me, I'm just worried about the static translator surviving this :D

src/DateComparison.php
gsteel: I can't get psalm to work with intersections of imported types, so that's why I repeated the declarations across every validator 😭

Ocramius: understandable :-)

src/AbstractValidator.php
gsteel: I decided to keep it for a small amount of BC - removing it will break translations where the validator composes an inner validator, for example `EmailAddress` uses the `Hostname` validator internally. Hopefully we can drop it in 4…

Ocramius: Worth having for v4: `__construct` not interfaced, `::fromArray()` interfaced :D

gsteel: I'm hoping that for v4 we might be able to reach immutability

PR #361, opened 2024-07-11
Refactor `ValidatorChain`

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes
| QA            | yes

### Description

- Adds native parameter and return types
- Drops "fluent" returns from all relevant methods
- Adds optional plugin manager constructor argument - one day we'll get rid of set/get plugin manager…
- Marks `getPluginManager`, `setPluginManager` and `plugin` methods as internal to the \Laminas namespace
- Removed deprecated methods
- Improve test coverage
- Minor changes to the docs such as referencing only validators shipped by this lib (As opposed to I18n) and removing the notice about service manager dependency requirement

### DIFF
--- a/src/ValidatorChain.php
+++ b/src/ValidatorChain.php
@@ -8,12 +8,12 @@ use Countable;
 use IteratorAggregate;
 use Laminas\ServiceManager\ServiceManager;
 use Laminas\Stdlib\PriorityQueue;
-use ReturnTypeWillChange;
 use Traversable;
 
 use function array_replace;
 use function assert;
 use function count;
+use function is_bool;
 use function rsort;
 
 use const SORT_NUMERIC;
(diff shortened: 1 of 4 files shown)


src/ValidatorChain.php
```diff
@@ -29,140 +29,107 @@ final class ValidatorChain implements Countable, IteratorAggregate, ValidatorInt
...
+    public function count(): int
     {
         return count($this->validators);
     }
 
     /**
      * Get plugin manager instance
      *
-     * @return ValidatorPluginManager
+     * @internal \Laminas
```
froschdesign, line 73: Can we have a better or more meaningful comment here?

froschdesign (changes requested):

src/ValidatorChain.php
gsteel: Yes! Thanks @froschdesign

Ocramius (approved):

src/ValidatorChain.php
froschdesign: And what is the meaning of `@internal \Laminas`?

gsteel: It scopes internal to a namespace so that, for example, input filter will not get 'internal method' psalm issues

froschdesign: The hint to [Psalm](https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-internal) helps here. Thanks! 👍🏻

froschdesign (approved):

PR #362, opened 2024-07-11
Remove the legacy `Upload` validator

### LINKED ISSUE #347: [RFC]: Deprecate and remove the `Upload` validator

gsteel:
Closes #347

See also #360


froschdesign (approved):
LGTM 👍🏻

PR #364, opened 2024-07-11
Return PHP_INT_MAX when a file size would exceed 2GB on a 32-bit system

### LINKED ISSUE #13: Error in function setMax from Zend\Validator\File\Size for values greater than 2G
If I'm using big values for the maximum limit, I have error on conversion in bytes.
In PHP, on a 32-bit system, the largest value that can be held in an INT is 2147483647. This is equal with 1.999999(and more digits) Gb.
I've tested the setMax with a greater value, like 8000M, but the tests can be made also with 2G.

The function setMax has a casting method to integer
`$max = (int) $this->fromByteString($max);`
This method will return a negative value because of the cast "(int)".

Example: setMax(7.81 GB) will have calculated max -204010947

The propose is to use "floor" casting like
`$max = floor($this->fromByteString($max));`

This bug is also presented in ZF3 package.

I'm giving you the example of usage of Size validator, from a getInputFilter() function. The value returned by the below call `$this->getUploadMaxFilesize()` is "8000M".

```php
           $file = new FileInput('imagepath');
            $file->getValidatorChain()->attach(new UploadFile());
            if($this->getUploadMaxFilesize()) {
                $file->getValidatorChain()->attach(new Size(array('max' => $this->getUploadMaxFilesize())));
            }
            $file->getValidatorChain()->attach(new Extension(array(
                'extension' => $this->getUploadFileExtensions(),
                'case' => false,
                'messages' => array('fileExtensionFalse' => "Only " . implode("/", $this->getUploadFileExtensions()) . " extension are allowed. File '%value%' hasn't an allowed extension"),
            )));
            $file->getFilterChain()->attach(
                new RenameUpload([
                'target' => PUBLICDIR . $this->getUploadPath() . $this->getUploadFilePrefix(),
                'randomize' => true,
                'use_upload_name' => false,
                'use_upload_extension' => true,
                'overwrite' => false,
            ]));
            $inputFilter->add($file);
```

---

Originally posted by @BogMor at https://github.com/zendframework/zend-validator/issues/199

gsteel:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes
| RFC           | yes

### Description

Closes #13 

File sizes > 2GB will return weird values on 32-bit systems as described by #13 

Should we consider this an issue at all?

If so, this patch will yield `PHP_INT_MAX` where it matters for file based validators.

### DIFF
--- a/src/File/Bytes.php
+++ b/src/File/Bytes.php
@@ -14,6 +14,7 @@ use function substr;
 use function trim;
 
 use const PHP_INT_MAX;
+use const PHP_INT_SIZE;
 
 /**
  * @internal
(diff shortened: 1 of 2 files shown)


Ocramius:
Are 32-bit systems still that relevant? :thinking:

gsteel:
> Are 32-bit systems still that relevant? 🤔

I didn't think so either, and #13 is pretty old - Happy for this to all go into the `wont-fix` pile

boesing:
7yo, I'd close it for now until there is a new report

Issue #370, opened 2024-07-25
New conditional validator


froschdesign:
The idea is to add a new validator that executes a validator when a rule is met.

## Example

```php
$validator = new Laminas\Validator\ConditionalValidator(
    [
        'validator' => [
            'name' => Laminas\Validator\NotEmpty::class,
        ],
        'rule'      => static function (array $context) {
            return (bool) ($context['confirm'] ?? false);
        },
    ]
);
```

```php
$data = [
    'registration' => 'foo',
    'confirm'      => '1',
];

var_dump($validator->isValid($data['registration'], $data)); // true

$data = [
    'registration' => '',
    'confirm'      => '1',
];

var_dump($validator->isValid($data['registration'], $data)); // false

$data = [
    'registration' => '',
    'confirm'      => '',
];

var_dump($validator->isValid($data['registration'], $data)); // true
```

## Implementation

The validator needs the plugin manager and a solution to allow a chain of validators, for example a third option – `validators` next to `validator`:

```php
$validator = new Laminas\Validator\ConditionalValidator(
    [
        'validators' => [
            [
                'name'    => Laminas\Validator\StringLength::class,
                'options' => [
                    'min' => 3,
                    'max' => 255,
                ],
            ],
            [
                'name' => Laminas\I18n\Validator\Alnum::class,
            ],
        ],
        'rule'      => static function (array $context) {
            return (bool) ($context['confirm'] ?? false);
        },
    ]
);
```

## Reference

https://book.cakephp.org/5/en/core-libraries/validation.html#conditional-validation

gsteel:
Nice addition for v3 👍

froschdesign:
@gsteel
Looks like we need the following logic from laminas-inputfilter then also here:

```php
    protected function populateValidators(ValidatorChain $chain, $validators)
    {
        foreach ($validators as $validator) {
            if ($validator instanceof ValidatorInterface) {
                $chain->attach($validator);
                continue;
            }

            /** @psalm-suppress RedundantConditionGivenDocblockType */
            if (is_array($validator)) {
                if (! isset($validator['name'])) {
                    throw new Exception\RuntimeException(
                        'Invalid validator specification provided; does not include "name" key'
                    );
                }
                $name    = $validator['name'];
                $options = [];
                if (isset($validator['options'])) {
                    $options = $validator['options'];
                }
                $breakChainOnFailure = false;
                if (isset($validator['break_chain_on_failure'])) {
[...]

PR #371, opened 2024-07-29
Add a simple factory for creating validator chains from configuration

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| New Feature   | yes

### Description

Adds a factory for creating validator chains from a well-known array shape.

The array shape is already in use in `laminas/laminas-inputfilter` and the factory introduced here can be used to to replace code in input-filter.

There's further background in #370 as this is a pre-requisite for introducing the proposed conditional validator.

### DIFF
--- a/docs/book/v3/validator-chains.md
+++ b/docs/book/v3/validator-chains.md
@@ -95,3 +95,50 @@ if ($validatorChain->isValid($username)) {
 
 // This first example will display: The input is less than 7 characters long
 ```
+
+## The Validator Chain Factory
+
+It is often desirable to create validator chains from configuration arrays.
+The `ValidatorChainFactory` enables this and expects configuration in the following shape:
+
+```php
+use Laminas\Validator\NotEmpty;
+use Laminas\Validator\StringLength;
+
+$chainConfiguration = [
+    'First' => [
+        'name' => NotEmpty::class,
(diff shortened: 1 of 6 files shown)


docs/book/v3/validator-chains.md
```diff
@@ -95,3 +95,48 @@ if ($validatorChain->isValid($username)) {
...
+
+It is often desirable to create validator chains from configuration arrays.
+The `ValidatorChainFactory` enables this and expects configuration in the following shape:
+
+```php
+use Laminas\Validator\NotEmpty;
+use Laminas\Validator\StringLength;
+
+$chainConfiguration = [
+    'First' => [
```
froschdesign, line 109: Is the key necessary here? (The same can be found in the tests.)

gsteel: The key is not necessary, but sometimes they are helpful for readability. The factory doesn't care if it's a list or a hash

froschdesign: > The factory doesn't care if it's a list or a hash

Right, so it should not be included in the example code, because the following text does not mention the array key either. Or it should be explained, e.g.:

> The factory does not use the key of the array so that it can be used by the user for better readability.

froschdesign (approved):
Nice! 👍🏻

PR #372, opened 2024-07-29
Add `EnumCase` and `BackedEnumValue` Validators

### LINKED ISSUE #167: Enum-Validator

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| New Feature   | yes

### Description

Adds 2 new validators to check that a given value is either a valid enum case, or a valid value for a backed enum.

Rather than expand the `InArray` validator, I prefer a more explicit approach which also keeps the `InArray` validator simple, and therefore:

- Closes #168
- Closes #167

### DIFF
--- a/docs/book/v3/set.md
+++ b/docs/book/v3/set.md
@@ -2,6 +2,7 @@
 
 The following validators come with the laminas-validator distribution.
 
+- [BackedEnumValue](validators/backed-enum-value.md)
 - [Barcode](validators/barcode.md)
 - [Callback](validators/callback.md)
 - [CreditCard](validators/credit-card.md)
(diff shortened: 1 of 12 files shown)


froschdesign (approved):
LGTM 👍🏻

PR #374, opened 2024-07-29
Remove outdated version indicators from validator docs

froschdesign:
Removed outdated "Since" version indicators from multiple validator documentation files to streamline the content. These lines were no longer necessary and cluttered the information provided about each validator's functionality.


gsteel (approved):
👍

Issue #96, opened 2021-03-07
Undocumented $context feature of validator's isValid($value, array $context = null) method


tempfirstuser:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

Why `$context` in validator's method  `isValid($value, array $context = null)` is undocumented feature? Does `$context` contain safe (filtered, validated) values to use them for query to database?
I needed validator for dependent fields, for, example, pair for username-token: token can be validated only if user exists and found that I can use $context besides 'callback' validator. I have found only one mentioned (just right now) in issues about `$context` https://github.com/laminas/laminas-validator/issues/15#issuecomment-570000069 by @weierophinney . 

Also, Metthew have answered to my question: "It's not documented, because it's not part of the validator interface. But the way validation chains work, they always get passed the context, which is the full set of values being validated. That said, they are the raw values, and there's no guarantee they are actually valid. The main purpose for having it passed is to do things like password verification (e.g., ensure the password in a reg form was typed twice identically), or to perform validations based on an aggregation of values (e.g., value is not valid by itself, but only when presented in combination with other values).".

PR #375, opened 2024-07-29
Introduce `Conditional` validator

### LINKED ISSUE #370: New conditional validator

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| New Feature   | yes

### Description

Closes #370

### DIFF
--- a/docs/book/v3/validators/conditional.md
+++ b/docs/book/v3/validators/conditional.md
@@ -1,0 +1,59 @@
+# Conditional Validator
+
+`Laminas\Validator\Conditional` allows you to validate a value conditionally depending on the outcome of a user-defined rule.
+
+## Supported options
+
+The following options are supported for `Laminas\Validator\Conditional`:
+
+- `rule`: A callable that will determine whether validation should execute on the value or not
+- `validators`: A list of validator specifications used to create a [validator chain](../validator-chains.md)
+
+## About the `$context` parameter of validators
+
+Typically, `laminas-validator` is used via `laminas-inputfilter` which is often, in turn, used via `laminas-form`.
+Some validators accept a second parameter to the `isValid()` method that contains the entire payload in an unfiltered and un-validated state.
+This parameter `$context` is normally the entire `$_POST` payload.
+
+`laminas-inputfilter` always passes this parameter to the `isValid` method, but, because it is not part of the `ValidatorInterface` contract, it's documentation has often been overlooked.
+
+## Basic usage
+
+The Conditional validator makes use of the `$context` parameter to determine whether validation should execute on the given value.
+
+As such, the `rule` parameter must be a callable that accepts the `$context` array and returns a boolean to indicate whether validation should proceed.
+
+If the rule returns false, the validator will return true, passing validation regardless of the input value. When the rule returns true, validation proceeds and returns either a successful or unsuccessful result depending on the input.
+
+The `Conditional` validator must create validation chains from the given `validators` option, therefore it requires the `ValidatorChainFactory` as a constructor dependency.
+
+Normally, using validators via `laminas-inputfilter` or `laminas-form`, this will not be a concern, but in standalone usage as documented here, we must account for this requirement.
+
+In the following example, we only wish to validate the email address if the `subscribe` field evaluates to a truthy value:
+
+```php
+use Laminas\Validator\Conditional;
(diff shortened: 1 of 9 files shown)


docs/book/v3/validators/conditional.md
```diff
@@ -0,0 +1,59 @@
...
+- `rule`: A callable that will determine whether validation should execute on the value or not
+- `validators`: A list of validator specifications used to create a [validator chain](../validator-chains.md)
+
+## About the `$context` parameter of validators
+
+Typically, `laminas-validator` is used via `laminas-inputfilter` which is often, in turn, used via `laminas-form`.
+Some validators accept a second parameter to the `isValid()` method that contains the entire payload in an unfiltered and un-validated state.
+This parameter `$context` is normally the entire `$_POST` payload.
+
+`laminas-inputfilter` always passes this parameter to the `isValid` method, but, because it is not part of the `ValidatorInterface` contract, it's documentation has often been overlooked.
```
froschdesign, line 18: This section is important and the parameter is also used in the following classes:

- Callback
- Explode
- Identical
- ValidatorChain

At least for the Callback validator, the section must be adopted, for all others it is internal processing.

See also: #96

docs/book/v3/validators/conditional.md
```diff
@@ -0,0 +1,59 @@
...
+Normally, using validators via `laminas-inputfilter` or `laminas-form`, this will not be a concern, but in standalone usage as documented here, we must account for this requirement.
+
+In the following example, we only wish to validate the email address if the `subscribe` field evaluates to a truthy value:
+
+```php
+use Laminas\Validator\Conditional;
+use Laminas\Validator\EmailAddress;
+use Laminas\Validator\ValidatorChainFactory;
+
+$chainFactory = $container->get(ValidatorChainFactory::class);
```
froschdesign: We already had this problem in the documentation of laminas-cache, so I propose to show the creation of the container:

```suggestion
$container = new Laminas\Validator\ValidatorPluginManager(
    new Laminas\ServiceManager\ServiceManager()
);
$chainFactory = $container->get(ValidatorChainFactory::class);
```

froschdesign (changes requested):

docs/book/v3/validators/conditional.md
gsteel: This has been documented in the Callback validator and CreditCard validator which composes a Callback validator in #379 … I'll do a separate patch for the other docs.

docs/book/v3/validators/conditional.md
gsteel: The chain factory is retrieved from the main DI container, not the plugin manger, so I've instantiated it directly in the docs instead.

froschdesign: Of course you're absolutely right!

froschdesign (approved):
LGTM 👍🏻

PR #253, opened 2024-06-12
Initial V3 Migration Guide

gsteel:
- Closes #251 
- Closes #292

PR #379, opened 2024-07-30
Firm up Callback arguments and document `$context` parameter.

gsteel:
Changes the Callback validator so that the callable always receives `(mixed $value, array $context, ...$otherArgs)` - this stabilizes the args for all environments, i.e. standalone usage and when part of input-filter etc.

Because the Callback validator is used internally, by the CreditCard validator, this also affects the signature of the `service` option to include context and declares the `$context` param in `CreditCard::isValid()` so that it can correctly be passed to the composed callback validator.

Will need to document this change in #253 

Edit: a bit more clarity…

Previously, the callback might receive `(mixed $value, ...$userDefinedArgs)` or `(mixed $value, array $context, ...$userDefinedArgs)` depending on whether the context was given to `Callback::isValid()` as the second param. This is difficult to document and hard to understand when the given arguments might differ.

This change means that the required signature of the callback is consistent at all times, but `$context` might be === [] in standalone usage.

### DIFF
--- a/docs/book/v3/validators/callback.md
+++ b/docs/book/v3/validators/callback.md
@@ -124,68 +124,59 @@ if ($valid->isValid($input)) {
 }
 ```
 
-## Adding Options
+## Validation Context Argument
 
-`Laminas\Validator\Callback` also allows the usage of options which are provided as additional arguments to the callback.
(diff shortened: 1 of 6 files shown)


gsteel:
Partially addresses #96

froschdesign (approved):
LGTM 👍🏻

PR #380, opened 2024-07-30
Add Support for PHP 8.4

gsteel:


Ocramius (approved):

Ocramius:
BTW, worth trying on 2.x?

.laminas-ci.json
```diff
@@ -1,6 +1,6 @@
 {
     "backwardCompatibilityCheck": true,
     "ignore_php_platform_requirements": {
-        "8.3": false
+        "8.4": true
```
froschdesign, line 4: This can be removed if all required components also support PHP in version 8.4?

gsteel: Once support is available in stdlib etc, yes but it's needed until then

froschdesign: Thanks for the clarification!

froschdesign: Do we have a process that checks this or do we have to go through everything manually later?

gsteel: We have to go through everything manually 😢

froschdesign (approved):

PR #383, opened 2024-07-31
Improve documentation of the `$context` parameter

### LINKED ISSUE #96: Undocumented $context feature of validator's isValid($value, array $context = null) method

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes

### Description

- Closes #96

### DIFF
--- a/docs/book/v3/validator-chains.md
+++ b/docs/book/v3/validator-chains.md
@@ -142,3 +142,13 @@ $factory = $container->get(ValidatorChainFactory::class);
 $chain = $factory->fromArray($chainConfiguration);
 $chain->isValid('Some Value');
 ```
+
+## About the `$context` Parameter
+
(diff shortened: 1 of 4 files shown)


froschdesign (approved):
A great improvement! 👍🏻

PR #384, opened 2024-07-31
Remove `laminas-modulemanager` support

### LINKED ISSUE #336: [RFC]: Drop Support for MVC in 3.0 so it can be added later without a BC break
## RFC
Validator ships a `Module` class without importing any symbols from MVC.

There are no tests for it and v3.0 of this lib will not work with any MVC project until _everything_ supports ServiceManager v4.x

### Proposal

Remove the Module class now, for a 3.0 release so that it can be reimplemented in a minor release, if necessary, in the future.

### Reasoning

By the time MVC supports SMv4, who knows what changes will need to occur to the Module class here, at which point we need to break BC on this class just so that MVC can be supported.

CC/ @laminas/technical-steering-committee

gsteel:
|    Q          |   A
|-------------- | ------
| BC Break      | yes

### Description

Removes module manager support in the validator plugin manager factory and in the shipped `Module` class.

Also removes `ValidatorProviderInterface` which is used exclusively in MVC related code.

Possibly closes #336


froschdesign (approved):
It makes sense to me. 👍🏻

PR #389, opened 2024-08-04
Introduce `DateIntervalString` validator

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| New Feature   | yes


froschdesign:
Should we provide a corresponding form element for this?

froschdesign (approved):
LGTM 👍

gsteel:
> Should we provide a corresponding form element for this?

Not sure about that. There's a _lot_ of potential config options, such as which units should be allowed, Y,M,W,D and Time. I use a custom HTML element (Jabbascript) to take care of that and create the interval spec in JS before populating a hidden element. I think that it would be very ugly and unweildy in plain markup / PHP if it were a series of number inputs with config options for "allow this" and "hide that"…

PR #390, opened 2024-09-10
Update Outdated Translation Information on Introduction Page 

gsteel:
|    Q          |   A
|-------------- | ------
| Documentation | yes

### Description

Intro page was outdated.

### DIFF
--- a/docs/book/v3/intro.md
+++ b/docs/book/v3/intro.md
@@ -143,80 +143,84 @@ if (! $validator->isValid('word')) {
 > $ composer require laminas/laminas-i18n
 > ```
 
-Validator classes provide a `setTranslator()` method with which you can specify
-an instance of `Laminas\Validator\Translator\TranslatorInterface` which will
-translate the messages in case of a validation failure. The `getTranslator()`
-method returns the translator instance. `Laminas\Mvc\I18n\Translator` provides an
-implementation compatible with the validator component.
+Validator classes provide a `setTranslator()` method with which you can specify an instance of `Laminas\Translator\TranslatorInterface` which will translate the messages in case of a validation failure.
+
+The `getTranslator()` method returns the translator instance.
 
 ```php
-use Laminas\Mvc\I18n\Translator;
+use Laminas\I18n\Translator\Translator;
 use Laminas\Validator\StringLength;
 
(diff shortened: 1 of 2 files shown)


docs/book/v3/intro.md
```diff
@@ -143,80 +143,84 @@ if (! $validator->isValid('word')) {
...
 > ```
 
-Validator classes provide a `setTranslator()` method with which you can specify
-an instance of `Laminas\Validator\Translator\TranslatorInterface` which will
-translate the messages in case of a validation failure. The `getTranslator()`
-method returns the translator instance. `Laminas\Mvc\I18n\Translator` provides an
-implementation compatible with the validator component.
+Validator classes provide a `setTranslator()` method with which you can specify an instance of `Laminas\Translator\TranslatorInterface` which will translate the messages in case of a validation failure.
+
+The `getTranslator()` method returns the translator instance.
```
froschdesign, line 148: The related `TranslatorAwareInterface` should not be in this component. At least this makes no sense to me here.

gsteel: TranslatorAwareInterface is used by the plugin manager to inject the dependency

froschdesign: Correct, but if we need the same thing in another component, it is created again?!

gsteel: We could add it to `Laminas\Translator` but I hate *Aware* interfaces - it's still setter injection.

If validator goes stateless, and we end up returning result instances - translation will end up being completely external, i.e:

```php

$result = $validator->validate($whatever);

if (! $result->isSuccess()) {
    echo $resultTranslator->translate($result);
}

```

… Or something like that…

I'm not sure that _now_ is the time to address it, unless you think it is, and we need more discussion!

froschdesign: > …but I hate Aware interfaces - it's still setter injection.

I've already thought about that.

> …translation will end up being completely external, i.e:

That should be the goal for version 4. We should therefore leave the interface in the current version (3) and remove it later.

_Thanks for the feedback! 👍🏻_

froschdesign (approved):
Good catch! 👍🏻

PR #396, opened 2024-10-21
`NotEmpty` Validator Input Filter Affordances

gsteel:
### Description

InputFilter needs access to the internal message templates of the `NotEmpty` validator in order to present error messages when a required field is empty. This patch re-introduces `getMessageTemplates` only on the NotEmpty validator and marks the method as internal.

https://github.com/laminas/laminas-inputfilter/blob/15e25a10db958c4f12b3eda19ae47f38d193eb58/src/Input.php#L457-L481

### DIFF
--- a/src/NotEmpty.php
+++ b/src/NotEmpty.php
@@ -260,4 +260,22 @@ final class NotEmpty extends AbstractValidator
 
         return true;
     }
+
+    /**
+     * Return the configured message templates
+     *
+     * This method is an affordance to laminas-inputfilter.
+     * It needs to introspect configured message templates in order to provide a default error message for empty inputs.
+     *
+     * In future versions of laminas-validator, this method will likely be deprecated and removed. Please avoid.
+     *
+     * @internal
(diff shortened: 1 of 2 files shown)


gsteel:
If I can get an approving review from @laminas/technical-steering-committee, then I think we can release v3.0 today 🤞

src/NotEmpty.php
```diff
@@ -260,4 +260,20 @@ public function isValid(mixed $value): bool
...
+
+    /**
+     * Return the configured message templates
+     *
+     * This method is an affordance to laminas-inputfilter.
+     * It needs to introspect configured message templates in order to provide a default error message for empty inputs.
+     *
+     * In future versions of laminas-validator, this method will likely be deprecated and removed. Please avoid.
+     *
+     * @psalm-internal \Laminas
```
Ocramius, line 274: Perhaps also `@internal` generally

Ocramius (approved):

PR #399, opened 2024-11-26
Fix options handling bug in Hash validator preventing the use of non-default algorithms

### LINKED ISSUE #398: Algorithm passed to Hash is only used if algorithm key is provided before hash key
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 2.x

#### Summary

The order of option keys determines whether the given algorithm is used.

#### Current behavior

If the algorithm key is passed after the hash key, the algorithm is silently set to crc32 (default).

#### How to reproduce

```php
// works
new Hash(['algorithm' => 'sha1', 'hash' => 'foo']);
// fails
new Hash(['hash' => 'foo', 'algorithm' => 'sha1']);
```

#### Expected behavior

Order of keys in options array is irrelevant.

InvisibleSmiley:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Fixes issue #398.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -3377,6 +3377,7 @@
       <code><![CDATA[getHash]]></code>
       <code><![CDATA[getHash]]></code>
       <code><![CDATA[getHash]]></code>
+      <code><![CDATA[getOptions]]></code>
       <code><![CDATA[setHash]]></code>
     </DeprecatedMethod>
     <MixedArgument>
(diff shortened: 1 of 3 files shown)


gsteel (approved):
Thanks @InvisibleSmiley :)

froschdesign:
Sorry, I added the wrong version for the bugfix. 🤦🏻‍♂️

PR #406, opened 2025-01-26
Upgrade Psalm to v6

gsteel:
Fixes a number of newly discovered Psalm issues mostly around possibly null or false return types from internal functions

Moves maglnet/composer-require-checker to its own CI job to prevent dependency conflicts

- Closes #407


bin/update_hostname_validator.php
```diff
@@ -115,11 +119,19 @@ function getNewValidTlds(string $string): array
 {
     // Get new TLDs from the list previously fetched
     $newValidTlds = [];
-    foreach (preg_grep('/^[^#]/', preg_split("#\r?\n#", $string)) as $line) {
+    $lines        = preg_split("#\r?\n#", $string);
+    assert(is_array($lines));
```
Ocramius: would be better to check that it isn't `false`?

.github/workflows/continuous-integration.yml
```diff
@@ -6,6 +6,27 @@ on:
...
+      - uses: actions/checkout@v4.2.2
+      - uses: shivammathur/setup-php@2.32.0
+
+        with:
+          php-version: ${{ env.default_php }}
+          ini-values: memory_limit=-1
+          extensions: ${{ env.php_extensions }}
+      - uses: ramsey/composer-install@3.0.0
+      - run: composer global config bin-dir /usr/local/bin
+      - run: composer global require -W maglnet/composer-require-checker
```
Ocramius: Eh, no, this is a recipe for destruction.

If you need to install the tool, then make sure it's locked in a `composer.json`/`composer.lock`

src/Barcode/Ean5.php
```diff
@@ -4,26 +4,32 @@
...
 namespace Laminas\Validator\Barcode;
 
+use Override;
+
 use function strlen;
 
 /** @psalm-import-type AllowedLength from AdapterInterface */
 final class Ean5 implements AdapterInterface
 {
+    #[Override]
```
Ocramius: I'd say disable the `#[Override]` rule, instead of adding it to the whole codebase.

`#[Override]` is (overall) a wasteful language addition, IMO

Ocramius (changes requested):

.github/workflows/continuous-integration.yml
gsteel: I don't understand - CRC is standalone - getting all our dev tools to play nice with deps is a pain and this just removes CRC from the possible conflicts?

gsteel: BTW, with Psalm at v6 on PHP 8.1, CRC is uninstallable

Ocramius: > getting all our dev tools to play nice with deps is a pain and this just removes CRC from the possible conflicts?

That is fine, but `composer global require -W maglnet/composer-require-checker` is equivalent to `"maglnet/composer-require-checker": "*"`, and that is NOT OK.

What you do to split out tools is creating a `composer.json` (**plus** `composer.lock`) in a subdirectory, and have every change committed too.

gsteel: OK. Gotcha! I've also added a Makefile to make all this a bit easier to manage in dev

Ocramius (approved):

Ocramius:
Thanks @gsteel!

PR #415, opened 2025-04-30
Use psalm prefix on phpdoc referring to psalm types

oldpec:
For maximum compatibility with other static analyzers that do not  understand @psalm-type, use @psalm- prefixed phpdoc when a psalm-type is referenced so that they do not consider the psalm-type a missing class.

|    Q          |   A
|-------------- | ------
| Documentation | yes
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

Currently the [Phan](https://github.com/phan/phan) static analyzer reports issues like:

`PhanTypeMismatchPropertyProbablyReal Assigning ($callback as a field) of type array<string,Closure(mixed,array|mixed=,mixed|string=):(bool|true)> to property but \ZF2Form\Validators\DependentValidate->options is \Laminas\Validator\OptionsProperty (no real type) (the inferred real assigned type has nothing in common with the declared phpdoc property type)`

Because it isn't aware of the meaning of:

```
* @psalm-type OptionsProperty = array{
 *     callback: callable|null,
 *     callbackOptions: array<array-key, mixed>,
 *     throwExceptions: bool,
 *     bind: bool,
 * }
 ```
so when Phan comes across:
```
    /**
     * Default options to set for the validator
     *
     * @var OptionsProperty
     */
    protected $options = [
```
Phan interprets 'OptionsProperty' as a type that doesn't exist.

Specifying @var and @psalm-var versions in this case should keep Psalm (and other static analyzers that understand @psalm* phpdoc) working optimally but also be compatible with other analyzers.

This change will help library users upgrade efficiently by not bothering them (or build systems) with false positives and allowing them to focus on real forwards compatibility issues

### DIFF
--- a/src/Callback.php
+++ b/src/Callback.php
@@ -64,7 +64,7 @@ class Callback extends AbstractValidator
 
     /**
      * @param array|callable $options
-     * @psalm-param OptionsArgument|callable $options 
+     * @psalm-param OptionsArgument|callable $options
      */
     public function __construct($options = null)
     {
--- a/src/Callback.php
+++ b/src/Callback.php
@@ -52,7 +52,8 @@ class Callback extends AbstractValidator
     /**
      * Default options to set for the validator
      *
-     * @var OptionsProperty
+     * @var array
+     * @psalm-var OptionsProperty
      */
     protected $options = [
         'callback'        => null, // Callback in a call_user_func format, string || array
@@ -61,7 +62,10 @@ class Callback extends AbstractValidator
         'bind'            => false, // Bind the callback to the validator instance
     ];
 
-    /** @param OptionsArgument|callable $options */
+    /** 
+     * @param array|callable $options
+     * @psalm-param OptionsArgument|callable $options 
+     */
     public function __construct($options = null)
     {
         if (is_callable($options)) {
--- a/src/Callback.php
+++ b/src/Callback.php
@@ -62,7 +62,7 @@ class Callback extends AbstractValidator
         'bind'            => false, // Bind the callback to the validator instance
     ];
 
-    /** 
+    /**
      * @param array|callable $options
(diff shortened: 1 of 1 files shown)


froschdesign:
@gsteel 
What do you think? For me is `@var array` too general and does not really help. But adding the half in `@var` and to full type in `@psalm-var` can not be the goal. Regardless of this, it affects the old version.

gsteel:
My initial reaction is that Psalm & PHPStan have no problem using type aliases in standard doc blocks such as `@var`, `@return` and `@param`, and, AFAIK both of these SA tools understand each others prefixes such as `@psalm-type` and `@phpstan-type`.

Amending doc blocks _(for the worse IMO)_ to support _another_ SA tool, with a smaller feature set and much smaller user/install base is not a good use of our limited maintainer time.

Reducing to `@param array` is a no from me, it would at minimum need to be `@param array<string, mixed>`, or preferably, `@param array{key: type, key2: type}`.

In the latter case, i.e the entire array shape, it would need to be duplicated in all positions which is exactly what type aliases help us avoid.

So, IMO, if your SA tool does not support type aliases, or, supports them but doesn't recognise the _(most popular)_ vendor prefixes during parsing, then that's a problem with the SA tool rather than something that should be fixed here.

Perhaps this is something we should discuss with the TSC?

cc @laminas/technical-steering-committee

Ocramius:
This has been an ever-present fight between using prefixed and non-prefixed docblock tags: we started moving away from `@psalm-`-prefixed blocks as PHPStan caught up with the type support, and I wouldn't go back for a single tool, at this point.

Closing here: Phan is currently not a focus, and it would be a massive game of whack-a-mole to support everything.

PR #430, opened 2025-08-06
Create and implement `ValidatorChainInterface` in final class `ValidatorChain`

### LINKED ISSUE #429: Final Validator chain without an interface - can't extend functionality
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

In my application I've created custom validator and filter chains because I had to implement some small modification in the validation logic. Now because of the new "trend", ValidatorChain and FilterChain classes got a final keyword, (which is questionable in my opinion, because Laminas is a set of packages, not a final application...). But I see this trend nowadays, and try to live with it. No matter if I know what I'm doing, I can't extend this classes. The recommended usual solution for this to decorate the classes, which can be a solution, if it has an interface. 

But these chain classes haven't any interface, and the final classes are used in a lot of dependencies in inputfilter package too (e.g. `setDefaultValidatorChain` method, constructors, etc.). And as I can see it doesn't planned in new release of inputfilter ( https://github.com/laminas/laminas-inputfilter/pull/144/files )

So my recommendation that if it is not a solution to not to mark classes as final, at least create an interface for it, and use it in dependencies.

If you approve it, I can create PRs for it

adambalint-srg:
Problem is came up in issue #429, final classes should have an interface.

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | yes
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

Create a ValidatorChainInterface, which extends ValidatorInterface (similar to filters). 
Also declare DEFAULT_PRIORITY in interface, but keep in ValidatorChain class for backward-compatibility, but with a deprecation note.

### DIFF
--- a/src/ValidatorChainFactory.php
+++ b/src/ValidatorChainFactory.php
@@ -16,7 +16,7 @@ final class ValidatorChainFactory
     }
 
     /** @param array<array-key, ValidatorSpecification> $specification */
-    public function fromArray(array $specification): ValidatorChainInterface
+    public function fromArray(array $specification): ValidatorChain
     {
         $chain = new ValidatorChain($this->pluginManager);
         foreach ($specification as $spec) {
--- a/src/ValidatorChainFactory.php
+++ b/src/ValidatorChainFactory.php
@@ -16,11 +16,11 @@ final class ValidatorChainFactory
     }
 
     /** @param array<array-key, ValidatorSpecification> $specification */
-    public function fromArray(array $specification): ValidatorChain
+    public function fromArray(array $specification): ValidatorChainInterface
     {
         $chain = new ValidatorChain($this->pluginManager);
         foreach ($specification as $spec) {
-            $priority   = $spec['priority'] ?? ValidatorChain::DEFAULT_PRIORITY;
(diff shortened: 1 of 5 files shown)


src/ValidatorChainFactory.php
```diff
@@ -16,11 +16,11 @@ public function __construct(private readonly ValidatorPluginManager $pluginManag
     }
 
     /** @param array<array-key, ValidatorSpecification> $specification */
-    public function fromArray(array $specification): ValidatorChain
+    public function fromArray(array $specification): ValidatorChainInterface
```
gsteel: The BC checker is correct here - the return type change should be reverted to `ValidatorChain`

gsteel (changes requested):
Thanks @adambalint-srg - just one trivial BC break to revert, otherwise LGTM 👍

src/ValidatorChainFactory.php
adambalint-srg: @gsteel Yes, I see now that this class is final too, so no extension will happen where interface would be useful. Reverted it

gsteel (approved):
Thanks @adambalint-srg 👍

PR #431, opened 2025-08-07
Create or Build `ValidatorChain` via `ValidatorPluginManager`

gsteel:
Whilst the V3 release included `ValidatorChainFactory` as a tool for correctly instantiating or building a chain, it neglegted the use-case of doing this directly via the plugin manager which is arguably a more common pattern.

Because `ValidatorChainFactory` already exists, this patch adds `ValidatorChainInvokableFactory` which delegates to `ValidatorChainFactory` because the latter has constructor arguments.

This patch is also relevant to work on `InputFilter` as we definitely need a way to do this via the plugin manager there.


ramchale (approved):

ramchale:
Looks good!

froschdesign (approved):

PR #437, opened 2025-08-14
Fix `ValidatorChainFactory` Building Plugins Manually

gsteel:
As per fix #435 - The factory now delegates to the chain when given a specification so that validators are either 'built' or 'got' depending on the presence of options so that `services` can be used in specifications.

Also adds runtime validation of the spec which could be considered a BC break 😬

### DIFF
--- a/src/ValidatorChainFactory.php
+++ b/src/ValidatorChainFactory.php
@@ -4,8 +4,6 @@ declare(strict_types=1);
 
 namespace Laminas\Validator;
 
-use function assert;
-
 /**
  * @psalm-import-type ValidatorSpecification from ValidatorInterface
  */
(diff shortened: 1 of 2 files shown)


gsteel:
Again, relevant to: https://github.com/laminas/laminas-inputfilter/pull/149

froschdesign:
@gsteel 
> Also adds runtime validation of the spec which could be considered a BC break 😬

What is the current behaviour without the checks if options are missing or wrong?

gsteel:
> What is the current behaviour without the checks if options are missing or wrong?

Type errors and php warnings

gsteel:
@froschdesign 

I wonder if it would be better to _add_ a static method to validator chain along the lines of
`ValidatorChain::assertSpecification(mixed $input): array` and then perform the runtime validation in `input-filter` and revert the automatic validation here.

WDYT?

gsteel:
@froschdesign 
I've reverted the addition of runtime checks - I think it's best to introduce tools for validating the spec array for consumers in a feature release. For now, this patch is effectively the same as #435 specifically for the chain factory 👍

ramchale (approved):

froschdesign (approved):

PR #439, opened 2025-08-25
fix message

e-belair:
fix error message template


gsteel (approved):
Thanks @e-belair 👍

Slamdunk:
This needs to be released for https://github.com/laminas/laminas-i18n-resources/pull/40 to have effect

PR #448, opened 2025-10-13
Drop support for PHP 8.1, add support for PHP 8.5

gsteel:
Adding 8.5 support to the 2.x series because installation of v3 is not yet widespread

### DIFF
--- a/src/HostWithPublicIPv4Address.php
+++ b/src/HostWithPublicIPv4Address.php
@@ -105,10 +105,6 @@ final class HostWithPublicIPv4Address extends AbstractValidator
 
         $filterFlags = FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE;
         if (PHP_VERSION_ID >= 80200) {
-            /**
-             * @psalm-var int $filterFlags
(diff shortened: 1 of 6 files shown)


src/HostWithPublicIPv4Address.php
```diff
@@ -105,10 +105,6 @@ public function isValid(mixed $value): bool
 
         $filterFlags = FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE;
         if (PHP_VERSION_ID >= 80200) {
-            /**
```
Ocramius, line 108: The conditional right above this is always true

Ocramius (changes requested):

src/HostWithPublicIPv4Address.php
Ocramius: Thanks, @github, for not allowing arbitrary comments on the diff :roll_eyes:

gsteel: 👍 Thanks! Amended

Ocramius (approved):
:ship:  thanks @gsteel!

PR #454, opened 2025-12-11
Loosen the `context` parameter of `isValid` to `array<array-key, mixed>`

gsteel:
It's entirely possible that the wider validation context might include data with integer keys. Restricting keys to `string` just causes downstream static analysis issues.

I also need this for `laminas-inputfilter` :)

### DIFF
--- a/src/ValidatorChainInterface.php
+++ b/src/ValidatorChainInterface.php
@@ -45,7 +45,7 @@ interface ValidatorChainInterface extends ValidatorInterface
      *
      * Validators are run in the order in which they were added to the chain (FIFO).
      *
-     * @param array<string, mixed> $context Extra "context" to provide the validator
+     * @param array<array-key, mixed> $context Extra "context" to provide the validator
      */
     #[Override]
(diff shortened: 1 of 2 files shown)


src/ValidatorChainInterface.php
```diff
@@ -45,7 +45,7 @@ public function attachByName(
      *
      * Validators are run in the order in which they were added to the chain (FIFO).
      *
-     * @param array<string, mixed> $context Extra "context" to provide the validator
+     * @param array<array-key, mixed> $context Extra "context" to provide the validator
```
Ocramius, line 48: This kinda forces an API expansion though: potentially BC issue

gsteel: You're (always 😁) right… but I'd bet money that the only implementation of this interface is right here in this library. If it was the `ValidatorInterface`, I think that the change would be problematic, but I believe this is more likely to help than hinder.

froschdesign (approved):

PR #457, opened 2026-01-20
Introduce new `KeyExists` Validator

gsteel:
If this addition is deemed sufficently generic and useful, please don't merge as I'll add docs to the same PR 👍


src/ValidatorPluginManager.php
```diff
@@ -175,6 +176,7 @@ final class ValidatorPluginManager extends AbstractSingleInstancePluginManager
             'isinstanceof'                 => IsInstanceOf::class,
             'isInstanceOf'                 => IsInstanceOf::class,
             'IsInstanceOf'                 => IsInstanceOf::class,
+            'keyExists'                    => KeyExists::class,
```
froschdesign, line 179: Only this one variant is possible? (Personally, I only use the class names.)

froschdesign (approved):

src/ValidatorPluginManager.php
gsteel: I added just the one, most obvious alias because "tradition" I'd prefer to drop them all because in a validation context, who needs anything other than the FQCN?

If you'd prefer this to be `KeyExists` (or to add the additional one), I'll happily change it!

Ocramius (approved):
Idea seems good to me! 

Would probably want a docs addition too, though

docs/book/v3/validators/key-exists.md
```diff
@@ -0,0 +1,73 @@
+# KeyExists Validator
+
+This validator allows you to validate if a given key is present in an iterable.
+
+## Supported Options
+
+The following options are supported for `Laminas\Validator\KeyExists`:
+
+- `key`: Sets the key that you wish to check.
```
froschdesign: I am unhappy with this section in general, but we need to rework it for all pages to make it simpler and clearer.

```suggestion
- `key`: Sets the key to be checked. `string` and `int` are supported.
```

docs/book/v3/validators/key-exists.md
```diff
@@ -0,0 +1,73 @@
...
+    'key' => '1.42',
+]);
+
+$validator->isValid(['1.0' => 'a', '1.42' => 'b']); // true
+```
+
+## Additional Notes
+
+Internally the validator calls [`iterator_to_array`](https://www.php.net/iterator_to_array) on the input to be validated.
+This may not be desirable if, for example, you are validating a queue where each item is dequeued during iteration, or there are other side effects of iteration on the subject.
```
froschdesign:
```suggestion
NOTE: **Additional Notes**
Internally the validator calls [`iterator_to_array`](https://www.php.net/iterator_to_array) on the input to be validated.
This may not be desirable if, for example, you are validating a queue where each item is dequeued during iteration, or there are other side effects of iteration on the subject.
```

froschdesign (changes requested):

froschdesign (approved):

src/ValidatorPluginManager.php
froschdesign: Maybe we should start a survey to find out what people actually want and use. The changes would then be implemented in the next major version.

PR #460, opened 2026-01-20
Add documentation for registering validators with the plugin manager

### LINKED ISSUE #95: Documentation missing section on registering custom validators and factories
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

I use apigility (yes, I didn't upgraded yet) for REST/RPC requests from ajax calls which means I configure everything in config files.
I just needed to create custom validator which will search in db if record exists with some conditions. I was trying to create service factory for that and pass service manager (which use doctrine) to validator `__construct` method, but it looks like doesn't work in such way. I was trying to do like this:

```
'service_manager' => [
	'factories' => [
	    Validator\EmailExistsAndConfirmed::class        => Validator\Factory\EmailExistsAndConfirmedFactory::class,
	],
],
```

but it doesn't work.

I didn't find anything on docs about that I should create validator factory and put it into config like this:

```
'validators' => [
	'factories' => [
	    Validator\EmailExistsAndConfirmed::class        => Validator\Factory\EmailExistsAndConfirmedFactory::class,
	],
],
```
It would be great to keep such hint in docs. If I missed this in docs, then sorry about that.

gsteel:
Closes #95

### DIFF
--- a/docs/book/v3/writing-validators.md
+++ b/docs/book/v3/writing-validators.md
@@ -275,17 +275,31 @@ The plugin manager always creates a new instance, providing options to the const
 
 When your validator has runtime dependencies on services, consider allowing an options array in the constructor so that the `AbstractValidator` options can be provided if required:
 
+#### Example Validator with a Service Dependency
+
 ```php
 namespace MyValid;
 
 use Laminas\Validator\AbstractValidator;
 use Psr\Container\ContainerInterface;
 
+/**
+ * @psalm-type Options = array{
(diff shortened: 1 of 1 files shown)


gsteel:
Hey @froschdesign - Can I poke you for a review!

froschdesign (approved):
Looks good in the current state. So, go for it! 👍 
(I have to think about how we can provide code examples for different applications or stand-alone.)

PR #473, opened 2026-05-18
Permit installation of `laminas/laminas-translator@2.0.0`

gsteel:
A release of laminas-validator with support for v2 of laminas-translator will be required in the 3.x series in order to make it possible to make a release of `laminas-i18n-validator:1.0.0`.

This patch should not be merged as-is. We will need to remove the dev dependency prior to releasing this as a minor.


froschdesign (approved):
Go for it! 👍🏻

PR #476, opened 2026-07-06
Replace PHP_Codesniffer and Psalm with Mago

gsteel:
The baselines here are pretty large:

- SA: 428 issues: 142 errors, 256 warnings, 30 help messages
- Lint: 172 issues: 49 errors, 64 warnings, 59 help messages

This is a relatively low-effort PR to illustrate what's involved to make the switch.

Bear in mind the the psalm baseline was _nearly_ empty here.

For comparison, Mago's lint, analyse and format runs each take a second, compared to Psalm at 35 - 40 seconds on this codebase.


froschdesign:
@gsteel 
Shall we discuss this at the next meeting? This pull request would be a great example of that.

gsteel:
@froschdesign 
Just writing a patch for tonights meeting now. That's what this is for 👍

gsteel:
Mago jobs now run directly out of the CI matrix so there's no need to configure them in yaml. Just switch to mago and profit!

src/File/FileInformation.php
```diff
@@ -25,13 +25,13 @@ final class FileInformation
...
     public readonly bool $readable;
-    private string|null $mediaType;
+    private ?string $mediaType;
 
     private function __construct(
         public readonly string $path,
         public readonly ?string $clientFileName,
         public readonly ?string $clientMediaType,
-        private Bytes|null $size,
+        private ?Bytes $size,
```
Ocramius, line 34: I disagree with some of these, but they are mechanical changes that can be shuffled around with some mago settings

mago.toml
```diff
@@ -0,0 +1,108 @@
...
+preserve-breaking-member-access-chain-first-method-on-same-line = true
+preserve-breaking-argument-list = true
+preserve-breaking-array-like = true
+preserve-breaking-parameter-list = true
+preserve-breaking-attribute-list = true
+preserve-breaking-conditional-expression = true
+preserve-breaking-condition-expression = true
+preserve-redundant-logical-binary-expression-parentheses = true
+
+space-after-logical-not-unary-prefix-operator = true
```
Ocramius, line 31: I wonder if there's a way to reference a common ruleset somehow?

composer.json
```diff
@@ -45,12 +45,10 @@
...
     },
     "require-dev": {
-        "laminas/laminas-coding-standard": "^3.1.0",
         "laminas/laminas-diactoros": "^3.8.0",
         "laminas/laminas-translator": "^2.0",
-        "phpunit/phpunit": "^11.5.55",
-        "psalm/plugin-phpunit": "^0.19.7",
-        "vimeo/psalm": "^6.16.1"
+        "phpunit/phpunit": "^11.5.56",
+        "carthage-software/mago": "^1.47.4"
```
Ocramius, line 51: This works, but in my case, mago will not run on my host ( https://mastodon.social/@ocramius/117162823461606054 )

Maybe I can contribute static compilation to upstream, though

Ocramius (approved):
I'm fine with this experiment 👍

mago.toml
gsteel: My plan is to make a Laminas specific `wyrihaximus/makefiles` repo for common internal tooling, this would include a org-wide `mago.toml` which can be included in the repo config file with something like `extends = vendor/foo/mago.toml`.

I should probably do that first to simplify things here perhaps.

Good/Bad idea?

src/File/FileInformation.php
gsteel: Yes - I also prefer `Type|null` - I'll figure out some more settings to get the diff smaller

composer.json
gsteel: Pretty sure all tooling is run out of Docker though if you use `make`…

https://github.com/laminas/laminas-validator/pull/476/changes#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R60

Nope! This should really be `docker run {stuff} mago analyse`…

I haven't yet got into Nix so I don't have the chops to ship a flake!

